PR #123, opened 2014-01-09
Allow .coffee by default

jasonals:
Use gulpfile.coffee if coffee-script is found.

I know this is already documented but having to type "gulp task --require coffee-script" instead of "gulp task" each time you want to run a task is a bit much, an extra 24 keystrokes. Please think about it.

### DIFF
--- a/bin/gulp.js
+++ b/bin/gulp.js
@@ -19,6 +19,8 @@ var cliPkg = require('../package.json');
 
 var localBaseDir = process.cwd();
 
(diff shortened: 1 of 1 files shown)


nmccready:
Where is the actual commit that provided coffee-script support on 3.7?

phated:
Support is handled by the `liftoff` module.  I don't have time to go back through the commit history, but you are welcome to.

PR #507, opened 2014-06-06
duplication: minimist is already used in gulp-util

alexeyraspopov:


enzuguri:
I was about to open a PR for the same issue, given that the original PR was June 6th 2014, do you still have plans to remove `gulp-util` at this stage? You could also further remove your dependency on chalk, as this is in `gulp-util` as `.colors`

PR #769, opened 2014-11-14
Tell user the local version is used in case of version mismatch

backflip:
Judging from the code, it looks like the global version always calls the local one (https://github.com/gulpjs/gulp/blob/master/bin/gulp.js#L119). In case of a version mismatch, it might be helpful to point this out.


billymoon:
I came here looking for a way to make gulp use the local version when available, only to find out that is what it does. I think it would be great if that was communicated more clearly in the error message.

billymoon:
+!

backflip:
My coworkers agree, too. Without the note, you wonder whether you have to do something about this.

contra:
Me or @phated will take a reasonable looking PR

PR #847, opened 2015-01-02
Use gulp.watch instead of 'update' event

DavidBM:
I like to propose this recipe for watchify.

I think that is more aligned with other gulp recipes.

Using watchify 'update' method, you can not declare dependencies for the the bundle() function. With gulp.watch you can.

What do you think?

(Sorry for my English, I'm still learning!)

### DIFF
--- a/docs/recipes/fast-browserify-builds-with-watchify.md
+++ b/docs/recipes/fast-browserify-builds-with-watchify.md
@@ -19,19 +19,19 @@ var bundler = watchify(browserify('./src/index.js', watchify.args));
 // add any other browserify options or transforms here
 bundler.transform('brfs');
 
-gulp.task('js', bundle); // so you can run `gulp js` to build the file
-bundler.on('update', bundle); // on any dep update, runs the bundler
+gulp.task('bundle-js', function () {
+	return bundler.bundle()
+		.on('error', gutil.log.bind(gutil, 'Browserify Error'))
+		.pipe(source('bundle.js'))
+		// optional, remove if you dont want sourcemaps
+		  .pipe(buffer())
+		  .pipe(sourcemaps.init({loadMaps: true}))
+		  .pipe(sourcemaps.write('./'))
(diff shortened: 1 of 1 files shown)


contra:
watchify.update works differently - it traces dependencies and updates when those change too, even outside of the src folder

DavidBM:
Ok, then better to use the 'update' event.

One question: How I can declare a dependency on the bundle function?

gulp.start is deprecated (and is not part of the public API of gulp).

And an example:

In my proposed code: `gulp.task('bundle-js', function () {`

I can change it with: `gulp.task('bundle-js', ['lint'], function () {`

How I can do that?

DavidBM:
I have been searching a bit and this is my conclusion.

Using run-sequence until Gulp 4.0 is released and punting `bundler on update` in the `watch` task.

Now I can define dependencies for the 'bundle-js' task.

``` javascript
var gulp = require('gulp');
var gutil = require('gulp-util');
var sourcemaps = require('gulp-sourcemaps');
var source = require('vinyl-source-stream');
var buffer = require('vinyl-buffer');
var watchify = require('watchify');
var browserify = require('browserify');
var runSequence = require('run-sequence'); //Hack waiting for gulp 4.0 https://www.npmjs.com/package/run-sequence

var bundler = watchify(browserify('./src/index.js', watchify.args));
bundler.transform('brfs');

gulp.task('bundle-js', function () {
    return bundler.bundle()
        .pipe(source('bundle.js'))
          .pipe(buffer())
          .pipe(sourcemaps.init({loadMaps: true}))
          .pipe(sourcemaps.write('./'))
        //
        .pipe(gulp.dest('./dist'));
});

gulp.task('watchify', function () {
	bundler.on('update', function () {
		runSequence('bundle-js');
	});
});
```

contra:
Yep - gulp 4 is almost out. You can use it now, just use the 4.0 branch

PR #866, opened 2015-01-12
Removed syntax highlighting from file structure

bradcerasani:
Fenced code block used shell syntax highlighting, drawing attention to `test` unnecessarily:

screen shot 2015-01-11 at 19 00 58

Additionally, is there any advantage to declaring `var dir`? This works the same:

```js
var requireDir = require('require-dir');

requireDir('./tasks');
```


sindresorhus:
> Additionally, is there any advantage to declaring var dir? This works the same:

Just for consistency I guess.

PR #872, opened 2015-01-19
bundler should be hoisted in a task

gasp:
- bundler var should not be in global scope
- this is a rather gulpish way of defining tasks
- watchify.fullPaths is usually not used in browserify

### DIFF
--- a/docs/recipes/fast-browserify-builds-with-watchify.md
+++ b/docs/recipes/fast-browserify-builds-with-watchify.md
@@ -15,23 +15,31 @@ var buffer = require('vinyl-buffer');
 var watchify = require('watchify');
 var browserify = require('browserify');
 
-var bundler = watchify(browserify('./src/index.js', watchify.args));
-// add any other browserify options or transforms here
-bundler.transform('brfs');
+// bundle the javascripts
+gulp.task('bundle', function() {
+  var options = watchify.args;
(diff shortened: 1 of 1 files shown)


contra:
This will actually slow things down by reconstructing the watchify and browserify instance multiple times

gasp:
Does it ? `bundler` does only exist in its task context but the task is only called once and therefore constructed only once.

adding those 2 console logs and running `gulp bundle` is quite explicit

```
// bundle the javascripts
gulp.task('bundle', function() {
  var options = watchify.args;
  options.fullPaths = false;
  console.log("watchify browserify, bundler is being instanced");
  var bundler = watchify(browserify('./src/main.js', watchify.args));
  // bundler.transform('brfs');

  // on any dep update, runs the bundler
  bundler.on('update', function() {
    bundle();
  });

  function bundle() {
    console.log("browserify is bundling files");
    return bundler.bundle()
      // log errors if they happen
      .pipe(source('bundle.js'))
        .pipe(buffer())
        .pipe(sourcemaps.init({loadMaps: true})) // loads map from browserify file
        .pipe(sourcemaps.write('./')) // writes .map file
      //
      .pipe(gulp.dest('./build'))
      .pipe(connect.reload());
  }
  return bundle();
});
 ```

Or am I missing something ?

PR #875, opened 2015-01-22
getting-started: package.json reminder

tomByrer:
Might be forgotten in a rush to start a new project.


contra:
I don't think we need to explain how to use npm, it's kind of assumed that people will know how to use it before trying to use this stuff

phated:
Not to mention, it will be in the gulp book when it comes out. We will be linking to the book in the docs

sindresorhus:
Agreed. We're not telling you how to make a folder either.

PR #894, opened 2015-02-09
bump gulp-cli (v8flags)

imyelo:


sindresorhus:
Not needed.

https://nodesource.com/blog/semver-tilde-and-caret

PR #924, opened 2015-02-24
suggesting a cleaner way

asfktz:
found it on:
https://medium.com/@sogko/gulp-browserify-the-gulp-y-way-bb359b3f9623

### DIFF
--- a/docs/recipes/browserify-uglify-sourcemap.md
+++ b/docs/recipes/browserify-uglify-sourcemap.md
@@ -9,8 +9,7 @@ Browserify with transforms and full sourcemaps that resolve to the original indi
 
 var browserify = require('browserify');
 var gulp = require('gulp');
-var source = require('vinyl-source-stream');
-var buffer = require('vinyl-buffer');
+var transform = require('vinyl-transform');
 var uglify = require('gulp-uglify');
 var sourcemaps = require('gulp-sourcemaps');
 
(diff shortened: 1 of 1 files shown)


contra:
This is going to create a bundle for every js file in the folder, I don't think that is what users are expecting

asfktz:
Absolutely right. Fixed.

contra:
@phated Thoughts? This is a lot cleaner IMO

phated:
:+1: I love vinyl-transform. It's like @hughsk's successor to the other vinyl-* modules he made. Even buffers by default. I think this should be the defacto way we promote using existing modules. Can you add a note to the top referencing the blog post?

contra:
Cool I'm +1 to merge this. If you see any other docs using vinyl-buffer feel free to update those as well!

asfktz:
I tried to make it work with the watchify example but with not success, can't figure way.. 

```js
var gulp = require('gulp');
var $ = require('gulp-load-plugins')();
var gutil = require('gulp-util');
var sourcemaps = require('gulp-sourcemaps');
var transform = require('vinyl-transform');
var watchify = require('watchify');
var browserify = require('browserify');

var bundler = watchify(browserify('./src/app.js', watchify.args));
//bundler.transform('brfs');

gulp.task('js', bundle); // so you can run `gulp js` to build the file
bundler.on('update', function (aa) {
    console.log(aa);
    bundle();
}); // on any dep update, runs the bundler

function bundle() {
    var bundled = transform(function () {
        return bundler.bundle();
    });

    return bundled
        // log errors if they happen
        .pipe(source('bundle.js'))
        .pipe(buffer())
        .pipe(sourcemaps.init({
            loadMaps: true
        }))
            .pipe($.react({harmony: true}))
            //.pipe($.uglify())
        // loads map from browserify file
        .pipe(sourcemaps.write('./')) // writes .map file
        //
        .pipe(gulp.dest('./dist'));
}
```

moimikey:
@contra @asfktz that `getBundleName` named func should probably get 86ed too.

callumacrae:
Doesn't this read the JS file from disk twice?

TomFrost:
@callumacrae is dead on the nose. My compile time is greatly reduced by changing

    return gulp.src('./app.js')

to

    return gulp.src('./app.js', {read: false})

at his suggestion.  Resulting file is the same. My version: https://github.com/TechnologyAdvice/Jexl/blob/master/gulpfile.js

PR #928, opened 2015-02-26
install cli globally

suyash:


phated:
ummm, this is going to break everyone's stuff. i am reverting.

contra:
woops thought this was for the 4.0 branch, my bad

suyash:
Can someone please point out exactly the features that might break when using gulp-cli globally with gulp 3 locally

phated:
@suyash https://github.com/gulpjs/gulp-cli/issues/7 has all the context

heikki:
Specific example on how things would have broken: https://github.com/npm/npm/issues/7130#issuecomment-74400623

suyash:
Thanks for clearing that up, but my use case is simple, I have gulp-cli installed globally (and not gulp) and gulp 3.8.11 as a devDependency (and not gulp-cli), and everything seems to work fine. I just want to confirm that the only issue with using gulp this way is the bin conflict that arises if you want to install both gulp and gulp-cli as npm dependencies, but other than that, everything __should__ work, and __will__ work once gulp 4 comes out and I simply upgrade my dependency in package.json

phated:
@suyash correct, everything works if you do the dance of removing gulp from global *then* install gulp-cli globally and don't try to install them at the same time.  It's not something we want to promote until the blockers are dealt with.

suyash:
Got it ! Thanks :+1:

Issue #925, opened 2015-02-24
Document updated gulp.watch syntax for gulp 4.0


kincaidoneil:
I recently updated to the 4.0 branch, however, I'm experiencing an issue in which using a function as a callback to `gulp.watch` results in nothing being based through to the `event` variable, as I particularly use `event.path`.

Here's my code:
```
gulp.watch('app/index.html', function(event) {
	console.log(event);
	console.log(event.path);
});
```

Here's what the console returns after the file is updated:
```
[Function: done]
undefined
```

Even if I pass through `gulp.series('name_of_task_here')`, it passes through the same thing to the first parameter.

Is there anything I'm doing wrong/was this method deprecated in gulp 4.0? If so, is there a better alternative?

Thanks!

heikki:
https://github.com/gulpjs/gulp/blob/master/docs/API.md#tasks

phated:
The watch syntax is changing in 4.0. Since everything is based on function composition, the watch tasks receive an async `done` callback (and handles streams/promises/etc being returned).

If you need to access the event object, use:

```js
var watcher = gulp.watch('app/index.html');
watcher.on('change', function(event){

});
```

phated:
Changing title to "Document updated watch syntax"

contra:
Should be in the changelog on the 4.0 branch

phated:
[Done!](https://github.com/gulpjs/gulp/blob/4.0/docs/API.md#gulpwatchglob-opts-tasks) - Thanks @dinoboff

lkraav:
> watcher.on('change', function(event){

Is `watcher.on('add change', function(event){` also supposed to work? It's not firing for me, while `'change'` alone is.

phated:
I doubt it, the emitter is just a node EventEmitter as far as I am aware. Please review `chokidar` docs for event support.

Issue #841, opened 2014-12-28
Document incremental builds


contra:
vinyl-fs has support for a new since parameter to only spit out modified files. we should document how to use this (possibly with a registry?) with the new task system

boogerlad:
This sounds like a great idea. If I'm understanding this correctly, the common "clean" task will no longer have to be ran to clear out the destination folder?

dinoboff:
#937 document the `gulp.src` `since` option.

I can document the new api to get the last run once (phated/undertaker#9) once it lands in Gulp.

phated:
Thanks to the efforts by @dinoboff this is implemented and documented in the 4.0 branch.  Closing.

Issue #803, opened 2014-12-02
Gulp 4 Docs


demisx:
We currently use gulp 3 for all our build needs. I'd like to come up to speed on gulp 4 and potentially move some of our projects to it. I have a couple of questions:

1. When is the anticipated release for v4?
2. What's the best place to start reading about v4? I am not sure if docs on 4.0 branch are up to date.

Thanks.

johnpapa:
I'm interested in contributing to this. I'll be releasing a 4+ hour video series in Gulp in late Jan 2015 and would like to help write the v4 docs, migration and more recipes for v4 when its time

travm:
I'll be available to help as well.

phated:
Some places to start:

https://github.com/phated/async-done
https://github.com/phated/undertaker
https://github.com/phated/bach
and the vinyl-* repositories

These docs will need to be translated to gulp-specific docs since gulp4 is built on top of them.

contra:
Yeah, we need help updating the `docs` folder which has become quite the large task. I'm open to seriously trimming down the docs folder to the essentials, reorganizing, redoing entirely, etc. for the gulp release. I will be doing markdown -> html and adding some nice styles to put the docs on the main site. One thing we failed really bad at was putting all the docs in one place (there are so many sub-projects) which I want to address in the 4.0 release.

sochka:
Any updates? Gulp 4.1 according to milestone page should have been released but 4.0 is not released yet.

contra:
@sochka Milestone dates mean nothing, we are busy people - it will get released when it gets released. Use the 4.0 branch until then.

phated:
I do actually have an update.  @dinoboff has been knocking out quite a few of the remaining blockers and I am using 4.0 in my current project to see if I run into any major bugs.  That being said, @contra is right, it will be released when it is ready.  The best thing you can do is use the 4.0 branch and let us know of any problems.

contra:
Also of relevance to this issue: We are working on a new website that will put all of the docs, blogs, guides, tutorials, recipes, etc. from all of the repos across the gulp ecosystem (vinyl, vinyl-fs, glob-watcher, etc.) in one place. The site will be searchable, indexed on google, and pretty badass.

callumacrae:
yay! :smiley:

contra:
[...]

PR #937, opened 2015-03-01
Document the new gulp.task API.

dinoboff:
Also add the gulp.series and gulp.parallel APIs.

#803, #841 and #925

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -2,52 +2,65 @@
 
 ### gulp.src(globs[, options])
 
-Emits files matching provided glob or an array of globs. 
-Returns a [stream](http://nodejs.org/api/stream.html) of [Vinyl files](https://github.com/wearefractal/vinyl-fs) 
-that can be [piped](http://nodejs.org/api/stream.html#stream_readable_pipe_destination_options) 
-to plugins.
+Emits files matching provided glob or array of globs.
+Returns a [stream] of [Vinyl files] that can be [piped] to plugins.
 
-```javascript
+```js
 gulp.src('client/templates/*.jade')
   .pipe(jade())
   .pipe(minify())
   .pipe(gulp.dest('build/minified_templates'));
 ```
 
-`glob` refers to [node-glob syntax](https://github.com/isaacs/node-glob) or it can be a direct file path.
+`glob` refers to [node-glob syntax][node-glob] or it can be a direct file path.
 
 #### globs
 Type: `String` or `Array`
 
 Glob or array of globs to read.
 
+Note that globs are evaluated in order, which means this is possible:
+```
+// exclude every JS file that starts with a b except bad.js
+gulp.src(['*.js', '!b*.js', 'bad.js'])
+```
+
 #### options
 Type: `Object`
 
 Options to pass to [node-glob] through [glob-stream].
 
-gulp adds some additional options in addition to the [options supported by node-glob][node-glob documentation] and [glob-stream]:
+gulp adds some additional options in addition to the
+[options supported by node-glob][node-glob documentation] and [glob-stream]:
 
-#### options.buffer
+##### options.buffer
 Type: `Boolean`
+
 Default: `true`
 
-Setting this to `false` will return `file.contents` as a stream and not buffer files. This is useful when working with large files. **Note:** Plugins might not implement support for streams.
+Setting this to `false` will return `file.contents` as a stream and not 
+buffer files. This is useful when working with large files.
 
-#### options.read
+**Note:** Plugins might not implement support for streams.
+
+##### options.read
 Type: `Boolean`
+
 Default: `true`
 
-Setting this to `false` will return `file.contents` as null and not read the file at all.
+Setting this to `false` will return `file.contents` as null and not read 
+the file at all.
 
-#### options.base
+##### options.base
 Type: `String`
+
 Default: everything before a glob starts (see [glob2base])
 
 E.g., consider `somefile.js` in `client/js/somedir`:
 
 ```js
-gulp.src('client/js/**/*.js') // Matches 'client/js/somedir/somefile.js' and resolves `base` to `client/js/`
+// Matches 'client/js/somedir/somefile.js' and resolves `base` to `client/js/`
+gulp.src('client/js/**/*.js')
   .pipe(minify())
   .pipe(gulp.dest('build'));  // Writes 'build/somedir/somefile.js'
 
@@ -56,11 +69,27 @@ gulp.src('client/js/**/*.js', { base: 'client' })
   .pipe(gulp.dest('build'));  // Writes 'build/js/somedir/somefile.js'
 ```
 
+##### options.since
+Type: `Date` or `Number`
+
+Setting this to a Date or a time stamp will discard any file that have not been 
+modified since the time specified.
+
+##### options.passthrough
+Type: `Boolean`
+
+Default: `false`
+
+If true, it will create a duplex stream which passes items through and 
+emits globbed files.
+
+
 ### gulp.dest(path[, options])
 
-Can be piped to and it will write files. Re-emits all data passed to it so you can pipe to multiple folders.  Folders that don't exist will be created.
+Can be piped to and it will write files. Re-emits all data passed to it so you
+can pipe to multiple folders. Folders that don't exist will be created.
 
-```javascript
+```js
 gulp.src('./client/templates/*.jade')
   .pipe(jade())
   .pipe(gulp.dest('./build/templates'))
@@ -69,32 +98,84 @@ gulp.src('./client/templates/*.jade')
 ```
 
 The write path is calculated by appending the file relative path to the given
-destination directory. In turn, relative paths are calculated against the file base. 
-See `gulp.src` above for more info.
+destination directory. In turn, relative paths are calculated against 
+the file base. See `gulp.src` above for more info.
 
 #### path
 Type: `String` or `Function`
 
-The path (output folder) to write files to. Or a function that returns it, the function will be provided a [vinyl File instance](https://github.com/wearefractal/vinyl).
+The path (output folder) to write files to. Or a function that returns it,
+the function will be provided a [vinyl File instance].
 
 #### options
 Type: `Object`
 
-#### options.cwd
+##### options.cwd
 Type: `String`
+
 Default: `process.cwd()`
 
-`cwd` for the output folder, only has an effect if provided output folder is relative.
+`cwd` for the output folder, only has an effect if provided output folder is 
+relative.
 
-#### options.mode
+##### options.mode
+Type: `String` or `Number`
+
+Default: the mode of the input file (file.stat.mode) or the process mode 
+if the input file has no mode property.
+
+Octal permission specifying the mode the files should be created with: e.g. 
+`"0744"`, `0744` or `484` (`0744` in base 10).
+
+##### options.dirMode
+Type: `String` or `Number`
+
+Default: Default is the process mode.
+
+Octal permission specifying the mode the directory should be created with: e.g.
+`"0755"`, `0755` or `493` (`0755` in base 10).
+
+##### options.overwrite
+Type: `Boolean`
+
+Default: `true`
+
+Specify if existing files with the same path should be overwritten or not.
+
+
+### gulp.symlink(folder[, options])
+
+Functions exactly like `gulp.dest`, but will create symlinks instead of copying
+a directory.
+
+#### folder
+Type: `String` or `Function`
+
+A folder path or a function that receives in a file and returns a folder path.
+
+#### options
+Type: `Object`
+
+##### options.cwd
 Type: `String`
-Default: `0777`
 
-Octal permission string specifying mode for any folders that need to be created for output folder.
+Default: `process.cwd()`
 
-### gulp.task(name[, deps], fn)
+`cwd` for the output folder, only has an effect if provided output folder is 
+relative.
 
-Define a task using [Orchestrator].
+##### options.dirMode
+Type: `String` or `Number`
+
+Default: Default is the process mode.
+
+Octal permission specifying the mode the directory should be created with: e.g.
+`"0755"`, `0755` or `493` (`0755` in base 10).
+
+
+### gulp.task([name,] fn)
+
+Define a task using [Undertaker].
 
 ```js
 gulp.task('somename', function() {
@@ -104,109 +185,195 @@ gulp.task('somename', function() {
 
 #### name
 
-The name of the task. Tasks that you want to run from the command line should not have spaces in them.
+Optional, The name of the task. Tasks that you want to run from the command line
+should not have spaces in them.
 
-#### deps
-Type: `Array`
+If the name is not provided, the task will be named after the function 
+`name` attribute, set on any named function.
 
-An array of tasks to be executed and completed before your task will run.
+[Function.name] is not writable; it cannot be set or edited.
+It will be empty for anonymous functions:
 
 ```js
-gulp.task('mytask', ['array', 'of', 'task', 'names'], function() {
-  // Do stuff
-});
+function foo() {};
+foo.name === 'foo' // true
+
+var bar = function() {};
+bar.name === '' // true
+
+bar.name = 'bar'
+bar.name === '' // true
 ```
 
-**Note:** Are your tasks running before the dependencies are complete?  Make sure your dependency tasks are correctly using the async run hints: take in a callback or return a promise or event stream.
+You should either provide the task name or avoid anonymous functions.
 
 #### fn
 
-The function that performs the task's operations. Generally this takes the form of `gulp.src().pipe(someplugin())`.
+The function that performs the task's operations. Generally it takes this form:
+```
+gulp.task('somename', function() {
+  return gulp.src(['some/glob/**/*.ext']).pipe(someplugin());
+})
+```
 
-#### Async task support
+Gulp tasks are asynchronous and Gulp uses [async-done] to wait for the tasks' 
+completion. Tasks are called with a callback parameter to call to signal 
+completion. Alternatively, Task can return a stream, a promise, a child process
+or a RxJS observable to signal the end of the task.
 
-Tasks can be made asynchronous if its `fn` does one of the following:
+**Warning:** Sync tasks are not supported and your function will never complete
+if the one of the above strategies is not used to signal completion. However, 
+thrown errors will be caught by Gulp.
+
+#### Async support
 
 ##### Accept a callback
 
-```javascript
-// run a command in a shell
-var exec = require('child_process').exec;
-gulp.task('jekyll', function(cb) {
-  // build Jekyll
-  exec('jekyll build', function(err) {
-    if (err) return cb(err); // return error
-    cb(); // finished task
-  });
+```js
+var del = require('del');
+
+gulp.task('clean', function(done) {
+  del(['.build/'], done);
 });
 ```
+
+The callback accepts an optional `Error` object. If it receives an error,
+the task will fail.
 
 ##### Return a stream
 
 ```js
 gulp.task('somename', function() {
-  var stream = gulp.src('client/**/*.js')
+  return gulp.src('client/**/*.js')
     .pipe(minify())
     .pipe(gulp.dest('build'));
-  return stream;
 });
 ```
 
 ##### Return a promise
 
-```javascript
-var Q = require('q');
+```js
+var Promise = require('promise');
+var del = require('del');
 
-gulp.task('somename', function() {
-  var deferred = Q.defer();
-
-  // do async stuff
-  setTimeout(function() {
-    deferred.resolve();
-  }, 1);
-
-  return deferred.promise;
+gulp.task('clean', function() {
+  return new Promise(function (resolve, reject) {
+    del(['.build/'], function(err) {
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
+  });
 });
 ```
 
-**Note:** By default, tasks run with maximum concurrency -- e.g. it launches all the tasks at once and waits for nothing. If you want to create a series where tasks run in a particular order, you need to do two things:
+or:
+```js
+var promisedDel = require('promised-del');
 
-- give it a hint to tell it when the task is done,
-- and give it a hint that a task depends on completion of another.
+gulp.task('clean', function() {
+  return promisedDel(['.build/']);
+});
+```
 
-For these examples, let's presume you have two tasks, "one" and "two" that you specifically want to run in this order:
-
-1. In task "one" you add a hint to tell it when the task is done.  Either take in a callback and call it when you're
-done or return a promise or stream that the engine should wait to resolve or end respectively.
-
-2. In task "two" you add a hint telling the engine that it depends on completion of the first task.
-
-So this example would look like this:
+##### Return a child process
 
 ```js
-var gulp = require('gulp');
-
-// takes in a callback so the engine knows when it'll be done
-gulp.task('one', function(cb) {
-    // do stuff -- async or otherwise
-    cb(err); // if err is not null and not undefined, the run will stop, and note that it failed
+gulp.task('clean', function() {
+  return spawn('rm', ['-rf', path.join(__dirname, 'build')]);
 });
 
-// identifies a dependent task must be complete before this one begins
-gulp.task('two', ['one'], function() {
-    // task 'one' is done now
-});
+```
 
-gulp.task('default', ['one', 'two']);
+##### Return a [RxJS] observable
+
+```js
+var Observable = require('rx').Observable;
+
+gulp.task('sometask', function() {
+  return Observable.return(42);
+});
 ```
 
 
-### gulp.watch(glob [, opts], tasks) or gulp.watch(glob [, opts, cb])
+### gulp.parallel(...tasks)
 
-Watch files and do something when a file changes. This always returns an EventEmitter that emits `change` events.
+Takes a number of task names or functions and returns a function of the composed
+tasks or functions.
+
+When using task names, the task should already be registered.
+
+When the returned function is executed, the tasks or functions will be executed
+in parallel, all being executed at the same time. If an error occurs, 
+all execution will complete.
+
+```js
+gulp.task('one', function(done) {
+  // do stuff
+  done();
+});
+
+gulp.task('two', function(done) {
+  // do stuff
+  done();
+});
+
+gulp.task('default', gulp.parallel('one', 'two', function(done) {
+  // do more stuff
+  done();
+}));
+```
+
+#### tasks
+Type: `Array`, `String` or `Function`
+
+A task name, a function or an array of either.
+
+
+### gulp.series(...tasks)
+
+Takes a number of task names or functions and returns a function of the composed
+tasks or functions.
+
+When using task names, the task should already be registered.
+
+When the returned function is executed, the tasks or functions will be executed
+in series, each waiting for the prior to finish. If an error occurs,
+execution will stop.
+
+```js
+gulp.task('one', function(done) {
+  // do stuff
+  done();
+});
+
+gulp.task('two', function(done) {
+  // do stuff
+  done();
+});
+
+gulp.task('default', gulp.series('one', 'two', function(done) {
+  // do more stuff
+  done();
+}));
+```
+
+#### tasks
+Type: `Array`, `String` or `Function`
+
+A task name, a function or an array of either.
+
 
 ### gulp.watch(glob[, opts], tasks)
 
+Watch files and do something when a file changes.
+
+```js
+gulp.watch('js/**/*.js', gulp.parallel('uglify', 'reload'));
+```
+
 #### glob
 Type: `String` or `Array`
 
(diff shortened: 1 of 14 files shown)


docs/API.md
```diff
@@ -104,132 +105,183 @@ gulp.task('somename', function() {
...
+});
 
-2. In task "two" you add a hint telling the engine that it depends on completion of the first task.
+```
 
-So this example would look like this:
+**Note:** By default, tasks run with maximum concurrency -- e.g. it launches 
+all the tasks at once and waits for nothing.
 
+To run serial tasks, use `gulp.serial` with asynchronous tasks:
```
heikki: gulp.series

docs/API.md
```diff
@@ -104,132 +105,183 @@ gulp.task('somename', function() {
...
+```js
+gulp.task('clean', function() {
+  return spawn('rm', ['-rf', path.join(__dirname, 'build')]);
+});
 
-2. In task "two" you add a hint telling the engine that it depends on completion of the first task.
+```
 
-So this example would look like this:
+**Note:** By default, tasks run with maximum concurrency -- e.g. it launches
```
heikki: Is this note still valid?

docs/API.md
```diff
@@ -104,132 +105,183 @@ gulp.task('somename', function() {
...
+}]));
+```
 
-#### opts
-Type: `Object`
+#### tasks
+Type: `Array`, `String` or `Function`
 
-Options, that are passed to [`gaze`](https://github.com/shama/gaze).
+A task name, a functions or an array of either.
```
heikki: function

docs/API.md
```diff
@@ -104,132 +105,183 @@ gulp.task('somename', function() {
...
 
-#### tasks
-Type: `Array`
 
-Names of task(s) to run when a file changes, added with `gulp.task()`
+### gulp.series(tasks, ...)
+
+Takes a variable amount of task names or function and returns a function
```
heikki: functions

docs/API.md
```diff
@@ -104,132 +105,183 @@ gulp.task('somename', function() {
...
+  // do more stuff
+  done();
+}]));
 ```
 
-### gulp.watch(glob[, opts, cb])
+#### tasks
+Type: `Array`, `String` or `Function`
+
+A task name, a functions or an array of either.
```
heikki: function

docs/API.md
dinoboff: You're right. It's not relevant in the `gulp.task` documentation

docs/API.md
```diff
@@ -92,9 +92,10 @@ Default: `0777`
 
 Octal permission string specifying mode for any folders that need to be created for output folder.
 
-### gulp.task(name[, deps], fn)
 
-Define a task using [Orchestrator].
+### gulp.task([name,] fn)
+
+Define a task using [Undertaker].
```
phated: should link to undertaker

docs/API.md
```diff
@@ -104,132 +105,161 @@ gulp.task('somename', function() {
...
 
-- give it a hint to tell it when the task is done,
-- and give it a hint that a task depends on completion of another.
+gulp.task('clean', function() {
+  return promisedDel(['.build/']);
+});
+```
 
-For these examples, let's presume you have two tasks, "one" and "two" that you specifically want to run in this order:
+##### Return a child process
```
phated: we also except RxJS observables now

docs/API.md
dinoboff: It is; it's a link and its URL is defined at the end of the file.

docs/API.md
```diff
@@ -104,132 +105,161 @@ gulp.task('somename', function() {
...
+var del = require('del');
 
-gulp.task('somename', function() {
-  var deferred = Q.defer();
+gulp.task('clean', function(done) {
+  del(['.build/'], function(err) {
+    done(err)
+  });
+  // or
+  // del(['.build/'], done);
```
phated: can we just use this example instead of the extra functions.  keep things more performant in the docs

docs/API.md
```diff
@@ -104,132 +105,161 @@ gulp.task('somename', function() {
...
-2. In task "two" you add a hint telling the engine that it depends on completion of the first task.
+```
 
-So this example would look like this:
 
-```js
-var gulp = require('gulp');
+### gulp.parallel(tasks, ...)
+
+Takes a variable amount of task names or function and returns
```
phated: need to mention that if you use task names, tasks must be registered prior to calling gulp.parallel (same goes for series)

docs/API.md
```diff
@@ -104,132 +105,161 @@ gulp.task('somename', function() {
...
+A task name, a function or an array of either.
 
-Watch files and do something when a file changes. This always returns an EventEmitter that emits `change` events.
 
-### gulp.watch(glob[, opts], tasks)
+### gulp.series(tasks, ...)
 
-#### glob
-Type: `String` or `Array`
+Takes a variable amount of task names or function and returns a function
```
phated: see note on parallel

docs/API.md
```diff
@@ -241,33 +271,26 @@ Type: `Object`
 
 Options, that are passed to [`gaze`](https://github.com/shama/gaze).
 
-#### cb(event)
-Type: `Function`
+#### tasks
+Type: `Array`, `Function` or `String`
 
-Callback to be called on each change.
+A task name, a function or an array of either to run when a file changes.
```
phated: make not about the reason to use `watcher.on('change')` if the gaze `event` is needed

docs/getting-started.md
```diff
@@ -3,7 +3,7 @@
 #### 1. Install gulp globally:
 
 ```sh
-$ npm install --global gulp
+$ npm install --global gulp-cli
```
phated: hold off on this until we get the blockers worked through

docs/API.md
```diff
@@ -104,132 +105,161 @@ gulp.task('somename', function() {
...
+bar.name = 'bar'
 ```
 
-**Note:** Are your tasks running before the dependencies are complete?  Make sure your dependency tasks are correctly using the async run hints: take in a callback or return a promise or event stream.
+You should either provide the task name or avoid anonymous functions.
 
 #### fn
 
-The function that performs the task's operations. Generally this takes the form of `gulp.src().pipe(someplugin())`.
+The function that performs the task's operations.
```
phated: note that these __cannot__ be synchronous anymore

docs/API.md
dinoboff: Do you have an example?

docs/API.md
phated: oh, okay

docs/API.md
phated: https://github.com/phated/async-done/blob/master/test/observables.js

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,65 @@
...
-
-For these examples, let's presume you have two tasks, "one" and "two" that you specifically want to run in this order:
-
-1. In task "one" you add a hint to tell it when the task is done. Either take in a callback and call it when you're done or return a promise or stream that the engine should wait to resolve or end respectively.
-
-2. In task "two" you add a hint telling the engine that it depends on completion of the first task.
-
-So this example would look like:
+- use `gulp.series`;
+- and your tasks should give hint of async operation by returning a stream,
```
phated: and RxJS observable

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,65 @@
...
+  // if err is not null and not undefined, the orchestration will stop,
+  // and 'two' will not run
+  done(err); 
 });
 
 // identifies a dependent task must be complete before this one begins
-gulp.task('two', ['one'], function() {
-    // task 'one' is done now
+gulp.task('two', gulp.series('one', function() {
+  // task 'one' is done now
```
phated: all tasks are async

phated:
@dinoboff you are amazing. thanks for all the hard work

docs/API.md
```diff
@@ -104,132 +105,157 @@ gulp.task('somename', function() {
 
 #### name
 
-The name of the task. Tasks that you want to run from the command line should not have spaces in them.
+Optional, The name of the task. Tasks that you want to run from the command line 
+should not have spaces in them.
 
-#### deps
-Type: `Array`
+If the name is not provided, the task will be name after the function `name` attribute.
```
contra: should note that named functions will have a .name attribute

docs/API.md
```diff
@@ -104,132 +105,157 @@ gulp.task('somename', function() {
...
+  done();
+});
 
-Options, that are passed to [`gaze`](https://github.com/shama/gaze).
+gulp.task('two', function(done) {
+  // do stuff
+  done();
+});
+
+gulp.task('default', gulp.series(['one', 'two', function(done) {
```
contra: are we going to show array or function arity usage in the docs? We have to pick one we can't use both. I vote for the variable function arity usage

docs/API.md
```diff
@@ -104,132 +105,157 @@ gulp.task('somename', function() {
...
+  // do more stuff
+  done();
+}]));
+```
 
 #### tasks
```
contra: should we really support this?

```js
gulp.watch('glob', ['one', 'two', 'three'])
```

Wouldn't it be easier to just tell people to provide a task fn as the second argument

```js
gulp.watch('glob', gulp.parallel('one', 'two', 'three'))
```

docs/recipes/running-tasks-in-series.md
dinoboff: So all tasks are async, they must either return async-y value use the callback argument?

docs/API.md
phated: @contra I believe we discussed this in the past and decided it wasn't worth breaking the syntax on watch and the array would just execute in parallel.  I remember rebasing to add the functionality back in after I removed.  I'd be fine with removing the array syntax if we want to change the decision (it simplifies .watch).

heikki: It would be good if `gulp.watch` and `gulp.task` had similar syntaxes.

docs/API.md
dinoboff: I am changing the note to:

    `Function.name` is not writable; it cannot be set or edited.
    It will be empty for anonymous function:

docs/API.md
```diff
@@ -104,132 +105,170 @@ gulp.task('somename', function() {
...
+var Observable = require('rx').Observable;
 
-// takes in a callback so the engine knows when it'll be done
-gulp.task('one', function(cb) {
-    // do stuff -- async or otherwise
-    cb(err); // if err is not null and not undefined, the run will stop, and note that it failed
+gulp.task('sometask', function() {
+  return Observable.return(42);
 });
+```
```
dinoboff: @phated is it ok?

docs/API.md
dinoboff: Can I leave that for #925 ?

docs/API.md
dinoboff: I was mixing them, It should now only show arity usage.

docs/recipes/running-tasks-in-series.md
phated: correct.  If you don't do either of those, the task hangs forever

docs/API.md
dinoboff: Changed. there shouldn't be any sync task left.

docs/API.md
```diff
@@ -104,132 +105,175 @@ gulp.task('somename', function() {
...
 
+The callback accept an optional `Error` object. If it receives an error, 
+the task will fail.
+
 ##### Return a stream
 
 ```js
 gulp.task('somename', function() {
-  var stream = gulp.src('client/**/*.js')
+  return gulp.src('client/**/*.js')
```
phated: can we make sure all examples have this update?  I like it much more than `var stream = ...; return stream`

dinoboff:
Thanks.

It covers `gulp.task` and `gulp.watch`. 

I don't know if there is any changes to `vinyl-fs` that need to be documented.

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,61 @@
...
 ```js
 var gulp = require('gulp');
 
 // takes in a callback so the engine knows when it'll be done
-gulp.task('one', function(cb) {
-    // do stuff -- async or otherwise
-    cb(err); // if err is not null and not undefined, the orchestration will stop, and 'two' will not run
+gulp.task('one', function(done) {
+  // do stuff -- async or otherwise
+  // if err is not null and not undefined, the orchestration will stop,
```
phated: remove the work "orchestration" throughout the docs

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,61 @@
...
+It also uses `once` to run dependencies once:
 ```js
 var gulp = require('gulp');
-var del = require('del'); // rm -rf
+var promisedDel = require('promised-del'); // rm -rf
+var once = require('once');
+
+gulp.task('clean', once(function() {
+  // The clean task will be run twice, but this function will be run once.
+  // So we cannnot rely on a callback.
```
phated: why can't we rely on the callback?  Is this a limitation of `once`?

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,61 @@
...
+  // Instead we return a promise; the same promise will be returned twice.
+  return promisedDel(['output']);
+}));
+
+gulp.task('templates', gulp.series('clean', function() {
+  var stream = gulp.src(['src/templates/*.hbs'])
```
phated: switch to the `return gulp.src` pattern

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,61 @@
...
-    del(['output'], cb);
 });
 
-gulp.task('templates', ['clean'], function() {
-    var stream = gulp.src(['src/templates/*.hbs'])
-        // do some concatenation, minification, etc.
-        .pipe(gulp.dest('output/templates/'));
-    return stream; // return the stream as the completion hint
+gulp.task('styles', gulp.series('clean', function() {
+  var stream = gulp.src(['src/styles/app.less'])
```
phated: see note above

docs/recipes/running-tasks-in-series.md
dinoboff: It was the only occurrence.

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,59 @@
...
-var del = require('del'); // rm -rf
+var promisedDel = require('promised-del'); // rm -rf
+var once = require('once');
 
-gulp.task('clean', function(cb) {
-    del(['output'], cb);
-});
+gulp.task('clean', once(function() {
+  // The clean task will be run twice, but this function will be run once.
+  // So we cannnot rely on a callback.
```
phated: why can't we rely on the callback? Is this a limitation of once? (added to the non-outdated diff)

docs/recipes/running-tasks-in-series.md
dinoboff: `once` will not run the function a second time, the callback will never be called and the task keep going on the second time it's called.

docs/recipes/running-tasks-in-series.md
phated: just dug into this and it seems to exist in both `once` and [`onetime`](https://github.com/sindresorhus/onetime) - I think we need an async version of `once` so it doesn't just rely on returning a promise

dinoboff: `once` will not run the function a second time, just send the return value of the first call for any other call.

Using the callback pattern, for the second call the callback will never be called and the task will keep going on.

phated: Just published a module that should solve that problem - https://www.npmjs.com/package/async-once

Please call that out as the solution to use with callbacks + an example

heikki: Is separate "Running tasks in series" recipe needed now that we have series / parallel? I'd also like to see all the references to "dependency" removed because there is no such thing anymore. Recipe for "Running function only once" might be useful for some, though I'd see that just as a sign of thinking the problem wrong way around.

phated: I think we could do away with the series recipe contents and just add a link that points to the docs for `series`.

I agree about doing away with the "dependency" terminology because everything is about function composition now.

I think we should have a "running once" recipe because people have come to expect that behavior from current gulp.  The recipe just makes the upgrade path nicer.

dinoboff: Updated:
- the example uses async-once.
- removed the dependency terminology, except that's I am calling the example using once a "dependency pattern".

callumacrae:
I can proofread this when it's done :smile:

dinoboff:
@callumacrae thanks; please do.

docs/API.md
```diff
@@ -2,46 +2,56 @@
 
 ### gulp.src(globs[, options])
 
-Emits files matching provided glob or an array of globs. 
-Returns a [stream](http://nodejs.org/api/stream.html) of [Vinyl files](https://github.com/wearefractal/vinyl-fs) 
-that can be [piped](http://nodejs.org/api/stream.html#stream_readable_pipe_destination_options) 
-to plugins.
+Emits files matching provided glob or an array of globs.
```
callumacrae: I would get rid of the "an" here, it makes it more obvious that the files are matching the array of globs, not that an array of globs is emitted.

docs/API.md
```diff
@@ -2,46 +2,56 @@
...
+```
+
 #### options
 Type: `Object`
 
 Options to pass to [node-glob] through [glob-stream].
 
-gulp adds some additional options in addition to the [options supported by node-glob][node-glob documentation] and [glob-stream]:
+gulp adds some additional options in addition to the
+[options supported by node-glob][node-glob documentation] and [glob-stream]:
```
callumacrae: I'm assuming the line breaks you're introducing into this file are to keep line lengths low?

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
+Optional, The name of the task. Tasks that you want to run from the command line
+should not have spaces in them.
 
-#### deps
-Type: `Array`
+If the name is not provided, the task will be name after the function `name` attribute.
 
-An array of tasks to be executed and completed before your task will run.
+`Function.name` is not writable; it cannot be set or edited.
+It will be empty for anonymous function:
```
callumacrae: functions plural

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
+You should either provide the task name or avoid anonymous functions.
 
 #### fn
 
-The function that performs the task's operations. Generally this takes the form of `gulp.src().pipe(someplugin())`.
+The function that performs the task's operations.
+Generally this takes the form of `return gulp.src().pipe(someplugin());`.
 
-#### Async task support
+Gulp assumes each task is asynchronous and call them with a callback parameter.
```
callumacrae: call -> calls

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
+A Task should either call the callback parameter to signal when operations
```
callumacrae: Task -> task, lowercase t

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
-  });
+```js
+var del = require('del');
+
+gulp.task('clean', function(done) {
+  del(['.build/'], done);
 });
 ```
 
+The callback accept an optional `Error` object. If it receives an error,
```
callumacrae: accept -> accepts

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
-  return deferred.promise;
+```js
+gulp.task('clean', function() {
+  return spawn('rm', ['-rf', path.join(__dirname, 'build')]);
 });
+
 ```
 
-**Note:** By default, tasks run with maximum concurrency -- e.g. it launches all the tasks at once and waits for nothing. If you want to create a series where tasks run in a particular order, you need to do two things:
+##### Return a RxJS observable
```
callumacrae: Is it work putting a link to RxJS here somewhere? Most people haven't heard of it.

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
+gulp.task('sometask', function() {
+  return Observable.return(42);
+});
+```
 
-For these examples, let's presume you have two tasks, "one" and "two" that you specifically want to run in this order:
 
-1. In task "one" you add a hint to tell it when the task is done.  Either take in a callback and call it when you're
-done or return a promise or stream that the engine should wait to resolve or end respectively.
+### gulp.parallel(tasks, ...)
```
callumacrae: Use a [rest parameter](http://tc39wiki.calculist.org/es6/rest-parameters/)! :smile: 

`gulp.parallel(...tasks)`

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
 
-2. In task "two" you add a hint telling the engine that it depends on completion of the first task.
+Takes a variable amount of task names or function and returns
+a function of the composed tasks or functions. When using task names,
+the task should already registered.
```
callumacrae: This doesn't make sense. Should already _be_ registered?

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
+A task name, a function or an array of either.
 
-Watch files and do something when a file changes. This always returns an EventEmitter that emits `change` events.
 
-### gulp.watch(glob[, opts], tasks)
+### gulp.series(tasks, ...)
 
-#### glob
-Type: `String` or `Array`
+Takes a variable amount of task names or function and returns
```
callumacrae: "Takes a number of task names" reads better, imo

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
+a function of the composed tasks or functions. When using task names,
+the tasks should already registered.
```
callumacrae: same again, missing a word here

docs/API.md
dinoboff: @phated the callback signature is `callback(error, result)` I think. Does the result parameter matter in Gulp?

docs/API.md
```diff
@@ -104,133 +177,185 @@ gulp.task('somename', function() {
...
 
 ```js
-var watcher = gulp.watch('js/**/*.js', ['uglify','reload']);
-watcher.on('change', function(event) {
-  console.log('File ' + event.path + ' was ' + event.type + ', running tasks...');
-});
+gulp.watch('js/**/*.js', ['uglify', 'reload', function(done) {
+  // do something
+  done();
+}]);
```
callumacrae: This isn't my decision to make, but I feel it probably shouldn't be encouraged to pass functions directly into `watch`. Sure, mention it below, but don't do it in a code example.

@phated @contra: thoughts?

docs/API.md
```diff
@@ -241,33 +366,47 @@ Type: `Object`
...
+gulp.watch('*.js', gulp.parallel(one, two));
+```
 
+`gulp.watch` returns an `EventEmitter` object which emits `change` events:
 ```js
-gulp.watch('js/**/*.js', function(event) {
+var watcher = gulp.watch('js/**/*.js', ['uglify', 'reload', function(done) {
+  // do something
+  done();
+}]);
```
callumacrae: I feel adding the anonymous function here adds needless complexity

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,95 @@
-# Running tasks in series, i.e. Task Dependency
+# Running tasks in series
 
-By default, tasks run with maximum concurrency -- e.g. it launches all the tasks at once and waits for nothing. If you want to create a series where tasks run in a particular order, you need to do two things:
+By default, gulp cli run task with maximum concurrency -- e.g. it launches all the tasks at once
```
callumacrae: cli -> CLI?

"--" -> "—"?

Why do you have a line break at 96 characters? lol

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,95 @@
...
+[`async-once`](https://www.npmjs.com/package/async-once) to run the `clean`
+task operations only once:
 ```js
 var gulp = require('gulp');
-
-// takes in a callback so the engine knows when it'll be done
-gulp.task('one', function(cb) {
-    // do stuff -- async or otherwise
-    cb(err); // if err is not null and not undefined, the orchestration will stop, and 'two' will not run
+var del = require('del'); // rm -rf
```
callumacrae: Why the `// rm -rf`?

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,95 @@
...
+var once = require('async-once');
+
+gulp.task('clean', once(function(done) {
+  // Run only once.
+  // For the next call to the clean task, once will call done with 
+  // the same arguments than the first call.
```
callumacrae: than -> as

callumacrae: also gulp uses lowercase letters at the beginning of comments, not sure if that's something you care about

docs/recipes/running-tasks-in-series.md
```diff
@@ -1,66 +1,95 @@
...
+// even though it is called as a dependency twice.
+gulp.task('build', gulp.parallel('templates', 'styles'));
 
-Another example, which returns the stream instead of using a callback:
+// an alias.
+gulp.task('default', gulp.parallel('build'));
+```
 
-```js
+Note that it's an anti-pattern in Gulp 4 and the logs will show the clean task
```
callumacrae: If it's an anti-pattern, why is it in the documentation?

callumacrae:
That's everything! Most of my comments can be ignored, but I hope some of them are helpful :smile:

docs/API.md
dinoboff: I agree. I put it there to show it like other composed task/function. But the main example be more explicit anyway, and use `gulp.parallel`:
```js
gulp.watch('js/**/*.js', gulp.parallel('uglify', 'reload'));
```
An example with array it shown later.

docs/recipes/running-tasks-in-series.md
dinoboff: I didn't know.

docs/recipes/running-tasks-in-series.md
dinoboff: I might be better to put it in the upcoming migration guide.

docs/recipes/running-tasks-in-series.md
dinoboff: Updated; I checked, I think it's the only occurrence.

dinoboff:
Thanks, they all made sense.

docs/API.md
dinoboff: checked.

docs/API.md
dinoboff: Yes, I didn't try to keep all lines under 80. I only cut the very long ones.

callumacrae: It's a bit odd, imo, especially here:

https://github.com/gulpjs/gulp/pull/937/files#diff-96cc378dcdef180641176501f42183c7R286

Imo (and what I usually do) is either consistently break the lines all at 80 characters, or don't break them.

dinoboff:
The travis fail is a timeout error unrelated to the doc changes.

phated:
I'm hoping to review the vinyl-fs docs and other changes Friday night or this weekend.

docs/API.md
```diff
@@ -69,32 +98,83 @@ gulp.src('./client/templates/*.jade')
...
+`cwd` for the output folder, only has an effect if provided output folder is 
+relative.
+
+##### options.mode
+Type: `String`
+
+Default: the mode of the input file (file.stat.mode) or the process mode 
+if the input file has no mode property.
+
+Octal permission string specifying the mode the files should be created with.
```
phated: is this a string?  or is it an actual Octal? `"0777"` vs `0777`

docs/API.md
```diff
@@ -69,32 +98,83 @@ gulp.src('./client/templates/*.jade')
...
+if the input file has no mode property.
+
+Octal permission string specifying the mode the files should be created with. 
+
+##### options.dirMode
+Type: `String`
+
+Default: Default is the process mode.
+
+Octal permission string specifying the mode the directory should be
```
phated: same question as above

docs/API.md
```diff
@@ -69,32 +98,83 @@ gulp.src('./client/templates/*.jade')
...
+
+### gulp.symlink(folder[, options])
+
+Functions exactly like `gulp.dest`, but will create symlinks instead of copying
+a directory.
+
+#### folder
+Type: `String` or `Function`
+
+A folder path or a function that takes in a file and returns a folder path.
```
phated: change `takes` to `receives`

docs/API.md
```diff
@@ -104,133 +184,191 @@ gulp.task('somename', function() {
...
 ```
 
-**Note:** Are your tasks running before the dependencies are complete?  Make sure your dependency tasks are correctly using the async run hints: take in a callback or return a promise or event stream.
+You should either provide the task name or avoid anonymous functions.
 
 #### fn
 
-The function that performs the task's operations. Generally this takes the form of `gulp.src().pipe(someplugin())`.
+The function that performs the task's operations. Generally this takes the 
+form of `return gulp.src().pipe(someplugin());`.
```
phated: this is the function body, not the whole task function.  I think you should make this a full example block of a task function

docs/recipes/mocha-test-runner-with-gulp.md
```diff
@@ -35,6 +35,6 @@ gulp.task('mocha', function() {
 });
 
 gulp.task('watch-mocha', function() {
-    gulp.watch(['lib/**', 'test/**'], ['mocha']);
+    gulp.watch(['lib/**', 'test/**'], 'mocha');
```
phated: switch `'mocah'` to `gulp.series('mocha')` just to stay consistent

docs/API.md
dinoboff: Tested, and both work.

docs/API.md
dinoboff: Tested, and both work.

docs/API.md
dinoboff: Like:

    The function that performs the task's operations. Generally it takes this 
    form:
    ```
    gulp.task('someTask', function() {
      return gulp.src(['some/glob/**/*.ext']).pipe(someplugin());
    })
    ```

docs/API.md
dinoboff: @phated Should I reword it to:

##### options.mode
Type: `String` or `Number`

Default: the mode of the input file (file.stat.mode) or the process mode if the input file has no mode property.

Octal permission specifying the mode the files should be created with: e.g. `"0777"`, `0777` or `511` (`0777` in base 10).

docs/recipes/server-with-livereload-and-css-injection.md
```diff
@@ -64,7 +64,7 @@ gulp.task('serve', function() {
     }
   });
 
-  gulp.watch('scss/*.scss', ['sass']);
+  gulp.watch('scss/*.scss', 'sass');
```
phated: switch `'sass'` to `gulp.series('sass')`

docs/recipes/using-external-config-file.md
```diff
@@ -30,10 +30,11 @@ Beneficial because it's keeping tasks DRY and config.json can be used by another
 ###### `gulpfile.js`
 
 ```js
-// npm install --save-dev gulp gulp-uglify
+// npm install --save-dev gulp gulp-uglify merge-stream
 var gulp = require('gulp');
 var uglify = require('gulp-uglify');
 var config = require('./config.json');
+var merge = require('merge-stream');
```
phated: this should be `merge2`

docs/API.md
phated: yeah

docs/API.md
phated: cool, thanks

docs/API.md
phated: yeah, i think that would be more clear

phated:
After my comments are addressed, I am :+1:  on this

dinoboff:
@contra can review and merge it please?

docs/API.md
dinoboff: done

docs/API.md
dinoboff: done

docs/API.md
dinoboff: done

docs/recipes/mocha-test-runner-with-gulp.md
dinoboff: done and it should be consistent.

docs/recipes/server-with-livereload-and-css-injection.md
dinoboff: done

docs/recipes/using-external-config-file.md
dinoboff: done

docs/recipes/using-external-config-file.md
```diff
@@ -30,10 +30,11 @@ Beneficial because it's keeping tasks DRY and config.json can be used by another
 ###### `gulpfile.js`
 
 ```js
-// npm install --save-dev gulp gulp-uglify
+// npm install --save-dev gulp gulp-uglify merge-stream
```
phated: missed updating this to merge2

phated: just checked the amount of downloads on each. I'm going to switch it back to merge-stream

docs/recipes/using-external-config-file.md
```diff
@@ -30,10 +30,11 @@ Beneficial because it's keeping tasks DRY and config.json can be used by another
...
 var gulp = require('gulp');
 var uglify = require('gulp-uglify');
 var config = require('./config.json');
+var merge = require('merge2');
```
phated: going to switch this back to merge-stream based on download count

phated:
merged

Ashwell:
I have been playing with gulp 4.0, works great for me. In looking at the docs, a function `gulp.symlink` has been documented but apparently not implemented as it is undefined on the gulp instance. Looking at index.html it would seem that below line number 13 this line should be added:

```js
Gulp.prototype.symlink = vfs.symlink;
```

no?

phated:
@Ashwell good catch, just pushed the addition.

Ashwell:
YAY! I helped!

PR #950, opened 2015-03-10
The task was not running before the build

bettiolo:
Option A:
 - This PR will make the task run *before* any other task, if they are added to the `default` task.
Option B:
- The wording should omit 'before running our build' as it makes you think this task is going to be synchronous


callumacrae:
del doesn't return anything. use a callback :smile: 

https://github.com/sindresorhus/del/blob/master/index.js#L20-L57

callumacrae:
Have you tested this? I'm pretty sure it breaks everything, and it worked fine before.

sindresorhus:
What @callumacrae said.

PR #957, opened 2015-03-13
Made browserify source maps actually work

callumacrae:
They didn't work before I made this change.

https://github.com/substack/node-browserify#var-b--browserifyfiles-or-opts


sindresorhus:
Browserify says the current way should work: https://github.com/substack/node-browserify#var-b--browserifyfiles-or-opts

callumacrae:
That's the same link as mine. This bit is key:

> When `opts.debug` is true, add a source map inline to the end of the bundle. This makes debugging easier because you can see all the original files if you are in a modern enough browser.

It looks like it defaults to false.

sindresorhus:
Oh, right. That's a weird name for a Source Map option.

callumacrae:
Yeah, took me a while to find!

PR #959, opened 2015-03-15
Bring JSCS to Gulp

mrjoelkemp:
TODO:

- [x] Fix violations

### DIFF
--- a/test/watch.js
+++ b/test/watch.js
@@ -9,7 +9,7 @@ var path = require('path');
 var should = require('should');
 require('mocha');
 
-var outpath = path.join(__dirname, "./out-fixtures");
+var outpath = path.join(__dirname, './out-fixtures');
 
 describe('gulp', function() {
   describe('watch()', function() {
@@ -21,7 +21,9 @@ describe('gulp', function() {
 
     var writeTimeout = 125; // Wait for it to get to the filesystem
     var writeFileWait = function(name, content, cb) {
-      if (!cb) cb = function() {};
+      if (!cb) {
+        cb = function() {}
+      };
       setTimeout(function() {
         fs.writeFile(name, content, cb);
       }, writeTimeout);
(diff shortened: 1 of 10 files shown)


test/watch.js
```diff
@@ -21,13 +21,15 @@ describe('gulp', function() {
 
     var writeTimeout = 125; // Wait for it to get to the filesystem
     var writeFileWait = function(name, content, cb) {
-      if (!cb) cb = function() {};
+      if (!cb) { cb = function() {} };
```
contra: should be on a new line, not sure if there is a JSCS rule for disallowing one-liners like this

callumacrae: There is: http://jscs.info/rules.html#requireblocksonnewline

test/watch.js
```diff
@@ -21,13 +21,15 @@ describe('gulp', function() {
...
       setTimeout(function() {
         fs.writeFile(name, content, cb);
       }, writeTimeout);
     };
 
+    // jscs: disable maximumLineLength
```
callumacrae: Why not just disable this rule in the test files instead of writing it out every single time?

test/watch.js
```diff
@@ -126,10 +131,11 @@ describe('gulp', function() {
         }, timeout);
 
         // it works if it calls the task
-        var watcher = gulp.watch(tempFile, {debounceDelay:timeout/2}, [task1,task2]);
+        var watcher = gulp.watch(tempFile, {debounceDelay: timeout / 2},
+          [task1, task2]);
```
callumacrae: Imo it would be better to just move the object into a variable:

```
var config = {debounceDelay: timeout / 2};
var watcher = gulp.watch(tempFile, config, [task1, task2]);
```

callumacrae:
What about adding jscs to the travis build? You can do some crazy stuff with `env.matrix` so that it runs separately and doesn't kill the whole build when jscs fails, it'll show that it's just a code style error.

test/watch.js
contra: Agree with @callumacrae

mrjoelkemp:
@callumacrae The typical use of JSCS is to fail builds to force people to fix style issues. This PR adds jscs to the `lint` task, which runs with the travis build.

I'm fine with whatever you guys want to do.

callumacrae:
It would fail the build, but it would still run the tests. I'm not sure how to demonstrate it without breaking my own build at work, though.

mrjoelkemp:
Would putting mocha in front of npm run-script lint solve the problem?
Doesn't jshint already fail builds?

phated:
I think it's proper to fail the build on invalid code styles.  We don't merge code if it doesn't match the style anyway.

PR #963, opened 2015-03-17
Added allowEmpty docs

callumacrae:
I don't think this will cherry-pick onto the 4.0 branch, I'll send another PR.


phated:
Is this going to be available in gulp 3.x?

callumacrae:
It was released in a minor version of glob-stream, so I would assume so. @contra?

contra:
I think glob-stream 4.0 is only being used by vinyl-fs 1.0 which is only being used by 4.0, so no not available in 3.x. Lots of new features and some breaking changes in the vinyl-fs 1.0 release

callumacrae:
:+1:

Issue #962, opened 2015-03-16
Document plugin API


nalbion:
The [writing a plugin](https://github.com/gulpjs/gulp/tree/master/docs/writing-a-plugin) docs are very vague.  Could you please document the parameters and returns?

````
/**
 * Assumptions, strategy etc...
 * @param {*} paramsFromGulp (optional) any parameters or config that may be passed from a Gulp script to the plugin
 * @return {no idea} Promise? Vinly? Stream? ????
 */
module.exports = function(paramsFromGulp) {

    /**
     * Assumptions, strategy etc...
     * @param {Vinyl} file - but pretend I have little/no experience with Vinyl or Node
     * @param {string} enc - presumably "utf8" etc
     * @param {function(Error?, Vinyl);WhatDoesItReturn?}
     * @return Promise? Vinly? Stream? ????   Because a callback is passed in, do we need to return anything?
     */
    return through.obj(function(file, enc, cb) {
    }
};
````

nalbion:
Can you provide any hints regarding the acceptance tests mentioned in "What happens if I don't follow them?"?

The "What does a good plugin look like?" section should explain why the example provided is a good looking plugin.

phated:
All of this can be figured out by looking at the modules they reference. That being said, we are always open to documentation PRs.

nalbion:
...or you could add a couple of lines of documentation and make it easier for everybody.
I've looked at some other gulp plugins, but they all seem to have been written by people who already know what they are doing.

Rather than expect people to do mental gymnastics, inferring how it is supposed to work by reading multiple other code bases, the "how to write a Gulp plugin" docs should at the very least document the contract between Gulp and the plugins.

----- Reply message -----

nalbion:
If you can answer the questions, and/or correct any misinterpretations in my snippet then I'll happily submit it as a PR

----- Reply message -----

phated:
Code is pretty impossible to critique as a comment in an issue. Please submit a PR with these docs so we can critique inline as per the contribution guidelines. Thanks for the time you've put into making gulp better for newcomers.

PR #966, opened 2015-03-17
Document plugin AP

nalbion:
This is my initial draft of additional documentation as discussed in #962.  At the moment, a lot of this is me thinking "out loud", but it may be helpful to others.

### DIFF
--- a/docs/writing-a-plugin/README.md
+++ b/docs/writing-a-plugin/README.md
@@ -14,7 +14,7 @@ If you plan to create your own Gulp plugin, you will save time by reading the fu
 A gulp plugin always returns a stream in [object mode](http://nodejs.org/api/stream.html#stream_object_mode) that does the following:
 
 1. Takes in [vinyl File objects](http://github.com/wearefractal/vinyl)
-2. Outputs [vinyl File objects](http://github.com/wearefractal/vinyl)
+2. Outputs [vinyl File objects](http://github.com/wearefractal/vinyl) (via `transform.push()` and/or the plugin's callback function) 
 
 These are known as [transform streams](http://nodejs.org/api/stream.html#stream_class_stream_transform_1) 
 (also sometimes called through streams). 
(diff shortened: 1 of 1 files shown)


nalbion:
I think that the main thing that was throwing me is that a lot of the plugins call `return callback(err,data);` - but my plugin needs to be asynchronous, so having to return anything immediately is not possible, and it's not obvious what the return value is.

...Upon further investigation it seems that callback() returns void, so it's just a short way of writing:

```
  callback(err,data);
  return;
```

contra:
Nicely done - yeah callback can be called at any time so it doesn't matter if your code is asynchronous or synchronous, just call it when you are ready

nalbion:
When I execute my gulp plugin:

```js
gulp.task('test', function() {
  gulp.src('test.js')
         .pipe(require('my-plugin')())
         .pipe(gulp.dest('output.js'));
});
```

Gulp passes me a Buffer.  Under what circumstances would it pass a Stream?

nalbion:
Is

```js
   this.emit('error', new PluginError(PLUGIN_NAME, 'Streams not supported!'));
   callback();
   return;
```

equivalent to

```js
   callback(new PluginError(PLUGIN_NAME, 'Streams not supported!'));
   return;
```

?

contra:
@nalbion It will pass a stream when you say `{buffer: false}` to src, useful for dealing with large files (lets say a 4GB video file) when your computer only has 1gb of ram

nalbion:
If you set `src: 'http://example.com/testfile.js'` would it give you a Buffer or a Stream?

contra:
@nalbion That doesn't exist and wouldn't work...

nalbion:
Is

   this.emit('error', new PluginError(PLUGIN_NAME, 'Streams not supported!'));
   callback();
   return;
equivalent to

   callback(new PluginError(PLUGIN_NAME, 'Streams not supported!'));
   return;
?

contra:
@nalbion depends on what module you're using, but in through2 yes

phated:
Thanks for all the work on this.  Great guide!

PR #974, opened 2015-03-22
Update doc and change log regarding task name

### LINKED ISSUE #969: Using strings when registering with gulp.task
Re: this entry in the [changelog](https://github.com/gulpjs/gulp/blob/4502a9caead8de00d16765dc30c4e5f81e847d75/CHANGELOG.md):

> using strings when registering with `gulp.task` should only be done when you will call the task with the CLI

A) What is the point of this advice? Is it that you can omit the string (the `name` arg) to make a task *not* callable from the CLI?

There was a pretty confusing discussion about that [starting here](https://github.com/gulpjs/gulp/issues/355#issuecomment-49510427).

Will this be callable via the CLI (`gulp whatever`)?:

```js
gulp.task(function whatever () {});

// As opposed to
gulp.task('whatever', function () {});
```

B) The API documentation for `.task()` should discuss this, right?

C) It looks like it will be an error to do `gulp.task(function () { "I'm anonymous"; })`, but if so the documentation doesn't make that clear.

D) Is `displayName` meant to be part of the API or it's just for some internal use?

```js
var f = function () {};
f.displayName = 'whatever';
gulp.task(f);
```

dinoboff:
fix #969

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -190,7 +190,7 @@ Octal permission specifying the mode the directory should be created with: e.g.
 
 ### gulp.task([name,] fn)
 
-Define a task using [Undertaker].
+Define a task exposed to Gulp command line; inherited from [Undertaker].
 
 ```js
 gulp.task('somename', function() {
@@ -200,13 +200,14 @@ gulp.task('somename', function() {
 
 #### name
 
-Optional, The name of the task. Tasks that you want to run from the command line
-should not have spaces in them.
-
 If the name is not provided, the task will be named after the function 
-`name` attribute, set on any named function.
+`name` or `displayName` attribute. The name argument is required if `fn` has 
+empty `name` and `displayName` attributes. 
 
-[Function.name] is not writable; it cannot be set or edited.
+Since the task will be run from the command line, you should avoid using 
+spaces in them.
+
+**Note:** [Function.name] is not writable; it cannot be set or edited.
 It will be empty for anonymous functions:
 
 ```js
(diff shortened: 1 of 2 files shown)


docs/API.md
```diff
@@ -190,7 +190,7 @@ Octal permission specifying the mode the directory should be created with: e.g.
 
 ### gulp.task([name,] fn)
 
-Define a task using [Undertaker].
+Define a task exposed to Gulp command line; inherited from [Undertaker].
```
contra: not a fan of this change

phated: I changed it in https://github.com/gulpjs/gulp/commit/9875ec578cc7bda9947cefa0c2c5fd5c29d7696f

docs/API.md
```diff
@@ -200,13 +200,14 @@ gulp.task('somename', function() {
...
 
-Optional, The name of the task. Tasks that you want to run from the command line
-should not have spaces in them.
-
 If the name is not provided, the task will be named after the function 
-`name` attribute, set on any named function.
+`name` or `displayName` attribute. The name argument is required if `fn` has 
+empty `name` and `displayName` attributes. 
+
+Since the task will be run from the command line, you should avoid using
```
contra: can be run from the command line*

You can also run tasks by name using `gulp.series('a', 'b')` so the docs shouldnt be too focused on the command line

phated: i've updated that also

Issue #977, opened 2015-03-22
remove array and string usage from gulp.watch


contra:
```js
gulp.watch('glob', ['task1', 'task2'])
gulp.watch('glob', 'task2')
```

will have to be changed to

```js
gulp.watch('glob', gulp.parallel('task1', 'task2'))
gulp.watch('glob', gulp.task('task2'))
```

This makes it a lot clearer what is going on.

Lots of documentation is going to need to be updated.

Semi-related: I also think we should discourage use of a watch task and just have people do it like this

```js
gulp.task('a', fn);
gulp.task('b', fn);

gulp.watch('glob', gulp.parallel('a', 'b'));
```

and if they have a flag for when to watch it can be as simple as

```js
gulp.task('a', fn);
gulp.task('b', fn);

if (process.env.NODE_ENV !== 'production) {
  gulp.watch('glob', gulp.parallel('a', 'b'));
}
```

I feel like this would simplify a lot of things

phated:
:+1: on the discouraging of task watches, especially because you get all the benefit of returned streams and stuff now (with gulp.series/parallel)

dinoboff:
I don't think any doc example is using array to define the tasks gulp.watch should run. [The only case I can find is](https://github.com/gulpjs/gulp/blob/4.0/docs/API.md#tasks-2):

> When `tasks` is an array, the tasks will be run in parallel:
> ```
> gulp.watch('*.js', [one, two]);
> // is equivalent to
> gulp.watch('*.js', gulp.parallel(one, two));
> ```

Maybe it should mention that it's ease migration and is deprecated.

dinoboff:
About discouraging use of a watch, I don't follow. The README gulpfile example watch task is:
```js
gulp.task(watch);
[...]
function watch() {
  gulp.watch(paths.scripts, scripts);
  gulp.watch(paths.images, images);
}
```
What are suggesting instead?

phated:
@dinoboff we are going to remove the array syntax altogether since 4.0 allows breaking changes.

The README example would be switched to:

```js
if (process.env.NODE_ENV !== 'production) {
  gulp.watch(paths.scripts, scripts);
  gulp.watch(paths.images, images);
}
```

I actually prefer:

```js
if (process.argv.indexOf('watch') !== -1) {
  gulp.watch(paths.scripts, scripts);
  gulp.watch(paths.images, images);
}
```

@contra thoughts?

contra:
[...]

PR #983, opened 2015-03-25
remove array and string usage from gulp.watch

dinoboff:
- Update API doc.
- recipes are unchanged; they didn’t use task name or task array for watch

#977

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -458,13 +458,16 @@ Type: `Array`, `String` or `Function`
 A task name, a function or an array of either.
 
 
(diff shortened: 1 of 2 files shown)


callumacrae:
Might be useful to specify that `gulp.series` and `gulp.parallel` return functions.

dinoboff:
@callumacrae Their doc mention it:

> Takes a number of task names or functions and returns a function of the composed tasks or functions.

I added a description to the example.

contra:
@phated thoughts?

phated:
This looks good

Issue #981, opened 2015-03-23
Source maps don't work for "Fast browserify builds with watchify" recipe


airandfingers:
The [Fast browserify builds with watchify](https://github.com/gulpjs/gulp/blob/master/docs/recipes/fast-browserify-builds-with-watchify.md) recipe uses gulp-sourcemaps to create a source map, but the generated source map is for bundle.js, not for the original src/index.js file.

I got source maps to work for me by following [christianalfoni's post](http://christianalfoni.github.io/javascript/2014/08/15/react-js-workflow.html); namely, the debug: true argument to the browserify call replaces gulp-sourcemaps.

contra:
@airandfingers You can rename it to whatever you want for the sourcemaps, if you have a correction for the recipe send a PR

contra:
This does raise the issue with debug: true not existing in the options passed to browserify though

phated:
I've fixed this in https://github.com/gulpjs/gulp/commit/7176929f09905cc2e7a2bb93f53779b54c081536

PR #984, opened 2015-03-26
Fix source maps usage by replacing gulp-sourcemaps with debug:true brows...

airandfingers:
...erify option

Fixes https://github.com/gulpjs/gulp/issues/981

### DIFF
--- a/docs/recipes/fast-browserify-builds-with-watchify.md
+++ b/docs/recipes/fast-browserify-builds-with-watchify.md
@@ -9,32 +9,27 @@ Watchify doesn't have a gulp plugin, but it doesn't need one either: you can use
 ``` javascript
 var gulp = require('gulp');
 var gutil = require('gulp-util');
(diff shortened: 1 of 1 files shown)


contra:
No reason to get rid of gulp-sourcemaps, just the option change will fix this

terinjokes:
Furthermore, keeping the `gulp-sourcemaps` usage ensures gulp plugins down stream properly handle the source maps.

airandfingers:
I haven't tested this with gulp-sourcemaps, since I already had source maps working. Feel free to modify this (or tell me how to modify it).

phated:
We wanted to keep gulp-sourcemaps and stuff so I've rewritten the watchify recipe in https://github.com/gulpjs/gulp/commit/7176929f09905cc2e7a2bb93f53779b54c081536

I've also tested and verified it is working as expected.  Thanks for pushing this forward.

PR #988, opened 2015-03-27
Update CLI.md

sudhirravindramohan:
Added a section on sending flags for tasks.


callumacrae:
Lowercase F. Also, would you be able to contribute this as a recipe?

sudhirravindramohan:
Corrected the upper case F. Regarding the recipe, we will be authoring some gulp tasks this week and I can give a subsequent pull request later.

phated:
I think this is fine.  We should put the answer into a recipe though instead of linking externally.

PR #994, opened 2015-03-31
use `else if`, no need to check the file for each type.

lookfirst:


phated:
No. Our pattern is early return and avoid else. Also, else-if isn't a real construct in JS

lookfirst:
I understand and appreciate that there is a particular style here, but the code doesn't early return.

phated:
These guidelines need to be updated, as plugins should only check for what they support and passthrough on every other type, but that doesn't require else-if statements.  Another thing that I missed at first was calling back with an error if it didn't match any types, which is a very bad pattern that we don't want to promote.  If you are willing to send another PR with my above comments, I will happily review it.  Thanks.

lookfirst:
@phated Even more confusing is that the example of the same exact plugin on the next page is not the same as the one on the first page. https://github.com/gulpjs/gulp/blob/master/docs/writing-a-plugin/using-buffers.md

Specifically... is this line important or not?

```
    // make sure the file goes through the next gulp plugin
    this.push(file);
```

phated:
`this.push(file)` is the most important part, it tells the stream to keep pushing the files along.

lookfirst:
So the most important bit is missing from the section titled 'What does a good plugin look like?'

I'm also really confused about Buffers vs. Streams. How does the end user have control in gulp over whether or not one or the other is presented to the plugin? 

It feels like the two different pages talking about buffers and streams should be combined into a single page and plugin authors should just support both.

phated:
The document that you edited is using the newer `cb(null, file)` pattern and the file you linked uses `this.push(file)`.  They both do the same thing, but I believe it is desired to point everyone to use the `cb(null, file)` pattern.

Streams are rarely used and are only really needed if you are manipulating large files like video and you don't want a 2gb file buffered into memory.  The vinyl objects generated by `gulp.src` can be buffer, null, or stream typed.  The options given to `gulp.src` are what determine the type.  Plugin authors can choose to error on stream type if they choose, although I would suggest they just pass those files through.

contra:
@phated they should error not pass it through to avoid confusion over if it is actually working or not

PR #998, opened 2015-04-02
Create dont-repeat-yourself-gulpfiles.md

lookfirst:


contra:
I don't think this is a pattern we want to endorse

phated:
:-1: from me

tkellen:
Same. The original version of grunt-contrib from 2012 used this pattern and it was a mistake (versioning stuff like this is a nightmare).

lookfirst:
I'd like to understand, beyond versioning, why this is a bad pattern or at least please offer a suggestion on what the 'right' way is. Cause copy/pasting code between gulpfiles doesn't work either.

phated:
@lookfirst take a look at custom registries to solve this problem - I believe they are doc'ed for 4.0 now.

lookfirst:
@phated You're going to make me dig up the link to the documentation? I've searched 'this repository' on github for 'custom registries' and I don't see anything other than some mentions in closed issues.

https://github.com/gulpjs/gulp/search?q=custom+registries&type=Issues&utf8=%E2%9C%93

tkellen:
This pattern is fine for one person or team of extremely well coordinated people. This pattern breaks down horribly if you try to use it with the programming community at large (unless you control every module in the kitchen sink, and even then it's a nightmare).

Here is what you have to look forward to:

**User:** What about the module I use all the time that you didn't include? Why can't that be in the kitchen sink too?
**Maintainer:** I don't use that module.
**User:** But you have 50 other modules, and I use all of them, why can't you add just that one?
**Maintainer:** Okay fine, what's the harm?

*500 issues open referencing problems using a module you didn't author, and don't even use.*

**User:** Module X bumped minor or major version, can you update this please?
**Maintainer:** Again? I released yesterday! I guess it doesn't matter if my module is at version `1500.0.0`...
**User:** Whoa, I'm like 4 major version behind! Hey maintainer, what changed?!

tkellen:
One possible solution:
`gulp --gulpfile /path/to/shared/gulpfile.js --cwd /path/to/project1`
`gulp --gulpfile /path/to/shared/gulpfile.js --cwd /path/to/project2`
`gulp --gulpfile /path/to/shared/gulpfile.js --cwd /path/to/project3`

phated:
@lookfirst sorry about that, it seems we haven't pulled in the undertaker docs for custom registries. Here is a link to the current undertaker docs on them https://github.com/phated/undertaker#custom-registries

That being said, I think @tkellen has outlined the problem well and I've implemented a system that allows other solutions so I am going to close this.

lookfirst:
@tkellen I hear you, the versioning issue is an issue, but I'm going to look beyond that one at this point because a DRY gulpfile is more important to me.

One giant shared gulpfile doesn't solve the DRY issue at all. I still have a package.json full of devDependencies and I don't even know which ones I need between projects. It requires an additional script and / or set of documentation just to build up those command arguments. The gulpfile itself isn't easy to install and isn't versioned with the project either.

There is already have a recipe for splitting the gulpfile. That is effectively what I've done. I've just taken it one step further and made the split code centralized. I see the split gulpfile recipe as more of a disaster than what I'm proposing. [Look at this setup](https://github.com/airbnb/airpal/tree/master/src/main/resources/assets), it is full of globals, the code isn't reusable and it is confusing to follow.  I don't see how that can be a recommended approach.

tkellen:
@lookfirst You don't need a single giant shared gulpfile, you could have 500 shared gulpfiles running off a single set of dependencies using the invocation format I mentioned.

I don't think there is anything wrong with what you are doing if the intent is for this to be consumed by a small, well organized team. In that context, your solution is very nice for keeping things DRY.

If you view this idea through the lens of a someone learning gulp (which is the point of the recipes) I think it's clear that they would be learning something quite different than gulp itself using this. I would view that as a disservice.

Finally... all abstractions leak. I think it's pretty clear the gulp community is fighting to have the thinnest abstractions possible while still providing powerful composition, and that is reflected in everything we publicly endorse.

lookfirst:
@tkellen Thanks for unlocking this thread and for the well thought out response.

> "If you view this idea through the lens of a someone learning gulp (which is the point of the recipes) they would be learning something quite different than gulp itself, which I see as a disservice."

That is the single best argument I've heard so far against including this recipe. I totally concur with you on that matter.

tkellen:
Sure thing Jon! Thanks again for taking the time to share this with us!

PR #1000, opened 2015-04-03
docs(recipes): Browserify's undefined behavior is undefined.

terinjokes:
b.bundle has always been documented as returning a readable stream. The fact that some versions return a duplex stream is an implementation detail, and writing to it results in undefined behavior.


phated:
thanks @terinjokes - browserify is simply the worst

PR #1005, opened 2015-04-05
add missing gulp-sourcemaps in browserify recipe

azu:
This pull request added missing `var sourcemaps = require('gulp-sourcemaps');` in browserify recipe.


phated:
Thanks!

PR #1007, opened 2015-04-05
checking for watch parameters, fixes #1002

### LINKED ISSUE #1002: Better error message when using remove gulp.watch functionality
In the 4.0 branch, we removed the ability to use strings as watch tasks and it causes a cryptic error message `TypeError: Arguments to path.resolve must be strings` as noted in #999 

We need to make this change/error more apparent.

ddprrt:
This is the PR addressing #1002. I found out that we not only have to deal with strings, but also with arrays (they have been used in a lot of documentation, too). This check makes those four little lines I added extremely ugly and hard to read. Input on that one is appreciated (esp. from you, @phated).

Also, tests have been added.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,6 +13,12 @@ Gulp.prototype.src = vfs.src;
 Gulp.prototype.dest = vfs.dest;
 Gulp.prototype.symlink = vfs.symlink;
 Gulp.prototype.watch = function (glob, opt, task) {
+  if (typeof opt === 'string' || typeof task === 'string' ||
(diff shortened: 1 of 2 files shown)


ddprrt:
Also, I do see that this solution still calls an Error. Considering that Gulp totally improved it's error reporting, we also might look for something prettier. This functionality tough seems to be hidden somehwere in the lib.

phated:
@ddprrt we are actually using node's built-in `assert` module for the better errors in the underlying libs.

phated:
Thanks for tackling this.

PR #1024, opened 2015-04-17
Update using-multiple-sources-in-one-task.md

hiddentao:
* Added documentation on how to do order the merged stream

### DIFF
--- a/docs/recipes/using-multiple-sources-in-one-task.md
+++ b/docs/recipes/using-multiple-sources-in-one-task.md
@@ -17,6 +17,25 @@ gulp.task('test', function() {
 });
 ```
 
+To control the order in which the streams are output once merged:
+
+```js
+// npm install --save-dev gulp ordered-merge-stream
+
+var gulp = require('gulp');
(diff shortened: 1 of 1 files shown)


TrySound:
@hiddentao It can be realized without dependencies by native Promises.
```
gulp.task('name', function () {
  return Promise.all([
    gulp.src('src1'),
    gulp.src('src2')
  ]);
});
```

hiddentao:
@TrySound Don't really understand, please explain. `Promise.all` waits until all promises have finished executing - it has no implications regarding the order of execution. Plus, I'm really trying to merge two streams which both output files and I want to merge the streams (and thus process each file) in a specific order.

callumacrae:
Streams aren't promises. Also, node.js doesn't have native promises yet, and not that people use iojs yet.

phated:
We aren't supporting event-stream anymore, as it doesn't always use real streams.  `ordered-merge-stream` uses it as a dependency.  I'd be willing to support a PR like this with a better library suggested.

Issue #904, opened 2015-02-14
Bring JSCS to Gulp?


mrjoelkemp:
Open to a PR that integrates JSCS into the travis flow?

phated:
That would help with a lot of PR comments about style.

mrjoelkemp:
Cool @phated. Thanks for the reply. I'll put one together.

mrjoelkemp:
@phated I'm looking around and don't see a styleguide for the Gulp codebase. Any reference with specific rules as to how you like the code written? Otherwise, I'll just whip up a JSCS config with rules that conform to the current style of the code.

phated:
@mrjoelkemp It's mostly just @contra's preferences.  If you submit something that passes with the current code base, we can work from there.

mrjoelkemp:
Cool. Thanks.

contra:
The styleguide for all of the projects is felixge's node-style-guide https://github.com/felixge/node-style-guide not just my preferences :palm_tree:

contra:
We should incorporate the jshintrc and eslintrc from that repo ^

This is a pretty big job considering how many repos we have now

heikki:
Shortcut to node-jscs issue -> https://github.com/jscs-dev/node-jscs/issues/1071

phated:
@mrjoelkemp since the jscs PR was merged, would you be able to submit a PR for this now?

mrjoelkemp:
Hey @phated. I will put a PR together. Without a new release of jscs, I can't use the `node-style-guide` preset directly. However, I could just copy the guts of it into gulp's jscsrc and then do a follow up PR (once the new version of jscs hits) converting the big jscsrc into:

```js
{
  "preset": "node-style-guide"
}
```

I'll put that together.

baer:
Is there a reason not to consider ESLint instead? It can serve the same purpose but it's based on the ESTree, is extendable and is almost fully supporting ES6 by now.

mrjoelkemp:
I'll avoid the general argument of JSCS vs ESLint. In this case, gulp's source isn't in es6 (though JSCS also has support for that). Plus, JSCS is a better fit for the task of thoroughly enforcing a styleguide.

mikesherov:
@baer :

> It is succinct extendable and based off of ESTree
As is esprima, who started the estree initiative to begin with, and is what espree is based off of.

> I can fill the role of JSHint + JSCS
Depends on what you think JSCS does.

[...]

PR #1034, opened 2015-04-21
Use JSCS-preset `node-style-guide`

SimenB:
Ref https://github.com/gulpjs/gulp/issues/904#issuecomment-80772730

`1.12.0` has been released.

It fails on maximum characters in a line. Should I disable that rule, or fix the errors?


callumacrae:
I usually disable that rule in test directories

SimenB:
JSHint also validates line length, but it's easier to disable it in tests because of the `extend` option which is not available in JSCS (yet, https://github.com/jscs-dev/node-jscs/issues/1106)

.jscsrc
```diff
@@ -1,48 +1,4 @@
...
-  },
-  "requireTrailingComma": {
-      "ignoreSingleLine": true
-  },
-  "requireBlocksOnNewline": true,
-  "validateIndentation": 2,
-  "validateLineBreaks": "LF",
-  "validateQuoteMarks": "'"
+  "preset": "node-style-guide",
+  "maximumLineLength": null
```
SimenB: This validation fails on tests, and is also validated by JSHint. JSHint makes it easy to disable only for tests but kept for all other files

callumacrae: It's not that tricky, just put `// jscs:disable maximumLineLength` at the top of test files

SimenB: It's double validation anyways, and as JSHint is run before JSCS, it will never matter

phated: I'm fine with this

PR #1035, opened 2015-04-21
Run JSHint on tests

SimenB:
JSHint was not run on the tests.

The failure is the same as in #1034, too long lines. Fix it, or ignore it in tests?

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -14,6 +14,5 @@
   "strict": true,
   "undef": true,
   "unused": true,
-  "node": true,
-  "mocha": true
+  "node": true
 }
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -1,0 +1,5 @@
+{
+  "extends": "../.jshintrc",
+  "maxlen": null,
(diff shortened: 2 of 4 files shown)


.jshintrc
```diff
@@ -14,6 +14,5 @@
   "strict": true,
   "undef": true,
   "unused": true,
-  "node": true,
-  "mocha": true
```
SimenB: No need to have `mocha` globals for the whole project

callumacrae:
:+1:

test/.jshintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "../.jshintrc",
+  "maxlen": null,
```
SimenB: Or I can actually fix the violations. A few were names of tests though, might be weird with a line-break in them?

callumacrae: No, you shouldn't make test names less clear in the name of coding standards

stevemao:
:+1:

PR #1042, opened 2015-04-26
Update delete-files-folder.md

vernonk:
The original path seems to also remove the `dist/mobile` folder (at least in my trial), this update leaves that original dir intact.


contra:
@sindresorhus is this behavior expected in del? I wouldn't expect `del('folder/**')` to deleter `folder/` just the files inside it

callumacrae:
I thought `**` optionally matched directories, so would match itself? It makes sense to me, and looks like it would be a thing in the grep library anyway, not in del.

PR #1046, opened 2015-04-27
Add a hint for browserify transforms

grmlin:
I added browserify transforms in every call of `bundle()` and got increasing build times while running. 
Transformations have to be added once in the beginning as discussed in this question here: https://github.com/substack/watchify/issues/187#issuecomment-89687576


contra:
Not sure we need to link to that thread, just

```js
// add transformations here
// i.e. b.transform(coffeeify);
```

grmlin:
This will be fine, of course.

PR #1056, opened 2015-05-04
Coding standards fix in plugin guidelines

callumacrae:


callumacrae:
worst diff ever. it's a function declaration, not an expression

stevemao:
I wish there was a tool just to check the code in README. I always make code errors in REAME and I see others make them too :sob:

callumacrae:
It would be a nice addition to jshint—it would be useful for inline JS in HTML as well, if you're into that. Would massively decrease performance and be a huge thing to implement, though.

Probably something to discuss over there, not here, though!

PR #1064, opened 2015-05-10
Improve the docs

lisposter:
I am translating gulp docs to Simplified Chinese, and I just find these.

It's all work fine, but it can be more accurate. :P


csyisaiko:
what"s your name?

contra:
Awesome! You should also check out the spanish translation project and see how they did it

lisposter:
@contra Thanks, yes, I will try my best to do it

PR #1067, opened 2015-05-13
edit the wrong code in doc 'delete-files-folder.md'

soliury:
In the older doc, the way to delete file in pipeline is wrong .

So I change to the right way.

The right way to delete pipeline should like this:

```
var gulp = require('gulp');
var stripDebug = require('gulp-strip-debug'); // only as an example
var del = require('del');
var vinylPaths = require('vinyl-paths');

gulp.task('clean', function () {
    var vp=vinylPaths();

     gulp.src('1.js')
     .pipe(vp)
    .pipe(stripDebug())
    .pipe(gulp.dest('dist'))
    .on('end',function(){
        del(vp.paths);
    })
});

gulp.task('default', ['clean']); 

```


contra:
How is this any better than just using gulp-rimraf?

phated:
I don't even understand this recipe.  Can anyone explain to me what it does better than the description in the file?

soliury:
In the older doc, it will go wrong. I think it was that del have changed api

docs/recipes/delete-files-folder.md
```diff
@@ -70,13 +70,20 @@ var del = require('del');
 var vinylPaths = require('vinyl-paths');
 
 gulp.task('clean:tmp', function () {
+  var vp = vinylPaths();
   return gulp.src('tmp/*')
+    .pipe(vp)
     .pipe(stripDebug())
     .pipe(gulp.dest('dist'))
-    .pipe(vinylPaths(del));
```
soliury: after that, it will delete the file pipe to the `dist`, it won't delete the files in the `tmp`

stevemao:
@contra gulp-rimraf is deprecated in favour of del because there is no need to have a gulp wrapper.
@phated The demo is trying to illustrate how to delete temp files after it's being processed. In the example, we want to process files in the tmp/ folder and then remove them afterwards. 
@soliury I think this is a bug in vinyl-fs as the example itself looks fine to me. After `gulp.dest('dist')` the vinyl path changes to `dist` (I don't think this is the correct behaviour. vinyl path shouldn't be changed after `gulp.dest`). Instead of deleting the files in tmp/ it deletes them in dist/. 
CC @sindresorhus

soliury:
OK, I see, I just hear somebody were asking this bug in the doc example, and I have try, it is really wrong, so I give this PR. 

But in the way that in the gulp doc, using it will go wrong. I have test, after the `gulp.dest` the vinyl path have changed, so use the older delete way will not go well.

sindresorhus:
The recipe is correct. If the paths are changed it's a bug in `vinyl-fs`.

stevemao:
Did anyone open an issue at `vinyl-fs`?

sindresorhus:
Turns out I was wrong.

docs/recipes/delete-files-folder.md
```diff
@@ -70,13 +70,20 @@ var del = require('del');
 var vinylPaths = require('vinyl-paths');
 
 gulp.task('clean:tmp', function () {
+  var vp = vinylPaths();
   return gulp.src('tmp/*')
+    .pipe(vp)
```
sindresorhus: Why not just move `.pipe(vinylPaths(del));` here? It's read into the disk anyways. So we can just delete it right away.

soliury: We should del the the tmp after the dest.

soliury: I have test the all usecase , it is right. See [this](https://cnodejs.org/topic/5552fc48c62bb7483db55101).

sindresorhus: > We should del the the tmp after the dest.

That's not really answering my question.

stevemao: @soliury its correct. We're trying to simplify it more :smile:

soliury: OK, I  see, yes we can delete the file directly.  Can change to this:

```
gulp.task('clean:tmp', function () {
  return gulp.src('tmp/*')
    .pipe(stripDebug())
    .pipe(gulp.dest('dist'))
    .on('end', function(){
         del('tmp/*');
    });
});
```

sindresorhus: No, then you have to glob them again. Please see my first comment for a good solution.

stevemao:
@soliury 

Just change it to
```js
gulp.task('clean:tmp', function () {
  return gulp.src('tmp/*')
    .pipe(vinylPaths(del))
    .pipe(stripDebug())
    .pipe(gulp.dest('dist'));
});
```
@sindresorhus believes this is simpler. I personally think there is no difference :)

soliury:
@stevemao @sindresorhus  Sorry to delay, I have changed the code.

PR #1079, opened 2015-05-23
Update API.md

yanxyz:
Change "globs" in gulp.src(globs[, options]) to "glob" is for uniform, the description says: `glob` refers to node-glob syntax and gulp.watch(glob[, opts], fn) uses "glob"


phated:
We accept an array of globs, I think I prefer plural here

yanxyz:
ok, I will make another pr.

PR #1081, opened 2015-05-24
bump liftoff and interpret

tkellen:
I'm not sure if we should release this as a patch or a minor bump. On the one hand, I think it is patch--for the end user it has the same features, just slightly improved for some corner cases. On the other, it does represent a pretty significant change in how Liftoff manages autoloads, so maybe it warrants a minor.


phated:
It adds features in the sense that babel was added, etc. I need to review the previous interpret we were installing and verify no transpilers were removed. Thanks for all the awesome work on this

phated:
I've gone over the extensions and stuff and I think this is going to work.  Merging.  I'll ping @contra to see what this should be released as.

Issue #284, opened 2014-02-19
Is it possible to copy files into different folders within one task?


leftstick:
My case was done by `Grunt` as following:

``` JavaScript
copy: {
            dev: {
                files: [{
                    expand: true,
                    cwd: 'bower_components/',
                    src: ['underscore/underscore.js', 'jquery/jquery.js', 'angular/angular.js', 'angular-route/angular-route.js', 'ng-grid/ng-grid-2.0.8.debug.js'],
                    dest: 'build/libs/',
                    flatten: true
                }, {
                    expand: true,
                    cwd: 'bower_components/',
                    src: ['ng-grid/ng-grid.css', 'jquery.ui/themes/base/jquery.ui.theme.css'],
                    dest: 'build/css/',
                    flatten: true
                }, {
                    expand: true,
                    cwd: 'bower_components/jquery.ui/themes/base/',
                    src: ['images/*'],
                    dest: 'build/css/'
                }]
            }
        }
```

I am wondering if it is possible to do the same thing with `Gulp`? Cause i like the gulp's streaming style, but no idea how to do it?

leadingcode:
// Copy recursive folder and all files from ./src to ./build DIR
// if you have copy all subcategories - use **/*
  gulp.task('copy', function () {
    gulp.src([
      './src/assets/**/*',
      './src/*.php',
      './src/parts/**/*',
      './src/phpexcel/**/*',
      './src/presentation/**/*',
      './src/scripts/**/*',
      './src/*.manifest',
      ],{
       "base" : "./src" 
     })
    .pipe(gulp.dest('build'));

GuilhermeMoreira:
1

yelexin:
This is my solution:
```js
function copyPasswordFaasTemplate(cb) {
  Promise.all([
    new Promise(resolve => {
      src(path.resolve(__dirname, 'src', 'userpool', 'settings', 'password', 'template.js'))
        .pipe(dest(path.resolve(__dirname, 'dist', 'src', 'userpool', 'settings', 'password')))
        .on('end', resolve);
    }),
    new Promise(resolve => {
      src(path.resolve(__dirname, 'src', 'userpool', 'settings', 'password', 'pwd-faas-tpl.yml'))
        .pipe(dest(path.resolve(__dirname, 'dist', 'src', 'userpool', 'settings', 'password')))
        .on('end', resolve);
    }),
    new Promise(resolve => {
[...]

PR #1084, opened 2015-05-28
Created recipe to show how to copy files

gortok:
This addresses a common question I see with Gulp (with SEO Friendly terms) and also brought up in issue #284.

### DIFF
--- a/docs/recipes/copy-files-different-destinations.md
+++ b/docs/recipes/copy-files-different-destinations.md
@@ -1,0 +1,17 @@
+# Copy files to different destinations
+
+Using gulp, you can copy files from different sources to different destinations in one task.
+
+```js
+gulp.task('copy-files', function () {
+   gulp.src('src/html/robots.txt')
+      .pipe(gulp.dest('dist'))
+  gulp.src('src/images/favicon.*')
+      .pipe(gulp.dest('dist'));
+  gulp.src('src/images/**')
+      .pipe(gulp.dest('dist/images'));
+});
+```
+
+See also: [Using multiple sources in one task](https://github.com/gulpjs/gulp/blob/master/docs/recipes/using-multiple-sources-in-one-task.md)
+ 


callumacrae:
-1, this will say that it is finished before it is done. the streams should be combined if you're going to do this

gortok:
@callumacrae After your helpful comment, I looked up the stream-combiner2 as a way to do this; but unfortunately the following code errors whenever you try to pass more than one src function to it that writes to the same location with a `write after end` error.

```js
gulp.task('move-statics', function () {
  var combined = combiner.obj([
    gulp.src('static/html/robots.txt')
      .pipe(gulp.dest('dist')),
    gulp.src('static/images/favicon.*')
      .pipe(gulp.dest('dist')),
    gulp.src('static/images/**')
      .pipe(gulp.dest('dist/assets/images'))
  ]);
  return combined;
});
```

From what I see from your activity, you seem to know gulp a great deal better than I do; can you suggest an improvement or how I could implement your suggestion?

KapuzenSohn:
For copieng Files I use something like this:
``` javascript
gulp.task('copy:images', function() {
  gulp.src(['src/**/[i|I]mg/**/*.*', 'src/**/[i|I]mages/**/*.*']) // ['file one', 'file two', 'file three', '...']
  .pipe(gulp.dest('dist'));
});
```

gortok:
@KapuzenSohn That's really interesting, but it doesn't cover the specific use-case I'm targeting: when you want to copy files to different places.  Could you show me the preferred way to do that?  It doesn't appear like my initial approach is appropriate (h/t to @callumacrae).

KapuzenSohn:
Oh, I see. I would combine the ```static/html/robots.txt``` and ```static/images/favicon.*´`` to a single gulp.src like I have and the other one should stay like you currently have.

Tahts the only solution I currently have. I looked at my Gulpfiles but I don´t have a similar problem.

callumacrae:
@gortok Check out this recipe: https://github.com/gulpjs/gulp/blob/master/docs/recipes/using-multiple-sources-in-one-task.md

The thing you're using is one of the other recipes, so that you can handle errors from all plugins in a task with the same handler.

phated:
I don't think we need this, as it is addressed in the recipe @callumacrae linked and shows how to correctly signal task completion.  @gortok If you would like to add some description/context to the linked recipe, I'll take a look at that PR.

PR #1087, opened 2015-06-03
Recipe: Accumulate end of the streams, not start

tepez:
Currently, we accumulate the start of the streams. As a result, the task ends before later processing is finished.

In my case, the later processing takes 20 seconds so it's very apparent.

### DIFF
--- a/docs/recipes/make-stream-from-buffer.md
+++ b/docs/recipes/make-stream-from-buffer.md
@@ -95,7 +95,7 @@ gulp.task('write-versions', function() {
       stream.end();
     });
 
-    streamEnd = stream
+    streamEnd = streamEnd
     // transform the raw data into the stream, into a vinyl object/file
     .pipe(vinylBuffer())
     //.pipe(tap(function(file) { /* do something with the file contents here */ }))
--- a/docs/recipes/make-stream-from-buffer.md
+++ b/docs/recipes/make-stream-from-buffer.md
@@ -79,12 +79,13 @@ gulp.task('write-versions', function() {
   availableVersions.forEach(function(v) {
     // make a new stream with fake file name
     var stream = source('final.' + v);
+    
+    var streamEnd = stream;
+    
     // we load the data from the concatenated libs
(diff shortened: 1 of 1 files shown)


docs/recipes/make-stream-from-buffer.md
```diff
@@ -79,13 +79,14 @@ gulp.task('write-versions', function() {
   availableVersions.forEach(function(v) {
     // make a new stream with fake file name
     var stream = source('final.' + v);
+    
+    var streamEnd = stream;
```
jmm: This assignment is superfluous, isn't it? You could just declare `streamEnd`, or eschew it altogether and just do `stream = stream ...` on line 98.

tepez: The former is possible but I wanted to make streamEnd more apparent.
I think the later is not possible because `stream.end()` is called async so we have to keep that reference.

jmm: > I think the later is not possible because `stream.end()` is called async so we have to keep that reference.

Oh, sorry, my mistake; I didn't notice that

> The former is possible but I wanted to make streamEnd more apparent.

IMO the unused assignment on line 83 is just confusing.

jmm:
Is this part actually even necessary in this scenario?:

```js
process.nextTick(function() {
  // in the next process cycle, end the stream
  stream.end();
});
```

As opposed to just:

```js
// write the file contents to the stream
stream.end(fileContents);
```

docs/recipes/make-stream-from-buffer.md
tepez: @jmm I added an extra commit.
The assignment inline 83 is now being used.
The purpose here is to show that you should always pipe to the end of the stream.

jmm: @tepez :+1: I think this is better than the original version, although `var streamEnd; ... streamEnd = stream ...` would've made sense to me also.

PR #1094, opened 2015-06-09
Update broken URL

kauegimenes:


callumacrae:
Why not just fix the URL?

https://github.com/google/web-starter-kit/blob/master/gulpfile.babel.js

phated:
can you point to their `gulpfile.babel.js` instead?

kauegimenes:
Sorry, updated the url to https://github.com/google/web-starter-kit/blob/master/gulpfile.babel.js

PR #1108, opened 2015-06-19
Update README.md

caseyhoward:
Eliminate needing to use gulp.series with only one task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,21 +36,22 @@ var paths = {
 };
 
 /* Register some tasks to expose to the cli */
-gulp.task('build', gulp.series(
-  clean,
-  gulp.parallel(scripts, images)
-));
+gulp.task(build);
 gulp.task(clean);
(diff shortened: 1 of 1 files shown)


phated:
This example doesn't work

caseyhoward:
Ah, my bad. Sorry about that.

caseyhoward:
It works if you call build or assign the result of gulp.series to a variable directly. I'm not sure how much better that is than what's already there though.

On a side note, I noticed while testing this that if I have a task that runs the same function as my default task and I define it after my default task, when I run gulp without a task name it says "Starting 'whatever'" instead of "Starting 'default'". Is this the expected behavior?

caseyhoward:
For example:
```js
var build = gulp.series('build');
gulp.task('default', build);
gulp.task('a', build);
```

phated:
that's interesting.  not sure what would be causing that

PR #1121, opened 2015-06-24
Simplify browserify-transforms.md

mikemaccana:
Specify full name of output explicitly
Remove unused reactify code
Tighten

### DIFF
--- a/docs/recipes/browserify-transforms.md
+++ b/docs/recipes/browserify-transforms.md
@@ -16,18 +16,13 @@ var buffer = require('vinyl-buffer');
 var gutil = require('gulp-util');
 var uglify = require('gulp-uglify');
 var sourcemaps = require('gulp-sourcemaps');
-var reactify = require('reactify');
 
 gulp.task('javascript', function () {
-  // set up the browserify instance on a task basis
-  var b = browserify({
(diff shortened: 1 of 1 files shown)


callumacrae:
Pretty sure the reactify stuff was half the point of this fiel

mikemaccana:
@callumacrae Text at top doesn't mention reactify at all. Unless the purpose of the demo is changed - it's better to keep the example focused on browserify.

jmm:
@mikemaccana I think you've inadvertently revealed that there's no reason for this document to exist anymore. With these changes I think it's just [Browserify + Uglify2 with sourcemaps](https://github.com/gulpjs/gulp/blob/master/docs/recipes/browserify-uglify-sourcemap.md).

mikemaccana:
@jmm *looks* Yup.

jmm:
@callumacrae @mikemaccana The point of the reactify part was that this recipe used to show a broken technique using `vinyl-transform` and placement of the transform registration code was significant. Then that whole technique was eliminated from the document.

jmm:
@mikemaccana So I would suggest to instead PR to remove this document, or change the name and turn it into a completely minimal gulp+browserify recipe without source maps and uglify.

mikemaccana:
*nod* will do

PR #1136, opened 2015-07-04
Update server-with-livereload-and-css-injection.md

pensierinmusica:
Browsersync offers an updated API for injecting CSS, as shown in their docs: http://www.browsersync.io/docs/gulp/

### DIFF
--- a/docs/recipes/server-with-livereload-and-css-injection.md
+++ b/docs/recipes/server-with-livereload-and-css-injection.md
@@ -79,12 +79,11 @@ app/
 var gulp = require('gulp');
 var sass = require('gulp-ruby-sass');
 var browserSync = require('browser-sync');
-var reload = browserSync.reload;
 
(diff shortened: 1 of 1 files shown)


shinnn:
@shakyShane I'd like you to review this.

Or, this markdown file can be replaced with a simple link instead of having almost the same recipes in those two different projects.

callumacrae:
I think that's a no!

Honestly, I can't see the point in this recipe; it seems redundant with the instructions on the browser-sync website. Could just replace it with a link to their docs? Anyone else have thoughts?

phated:
I would love to offload any and all recipes to the docs in their respective libraries so I agree with @callumacrae - closing this in favor of replacing the recipe with a link to http://www.browsersync.io/docs/gulp/

shakyShane:
There's also a bunch of our own working recipes https://github.com/BrowserSync/recipes

PR #1152, opened 2015-07-14
Added plugin naming guidelines

### LINKED ISSUE #1058: Plugin naming convention
There's nothing in the guidelines about naming plugins. I'd suggest the following:

- If the sole purpose of the package is to be a gulp plugin, it should begin "gulp-".
- If it is not a gulp plugin (for example, if it is only gulpfriendly), it should not start "gulp-".
- It should not contain anything other than alphanumeric characters and hyphens.

callumacrae:
Closes #1058.


contra:
Perfect, thanks!

PR #1171, opened 2015-07-27
add support for .gulprc

stefanr:


tkellen:
Thanks @stefanr, but this change needs to happen in Liftoff. See tkellen/js-liftoff#8 for a proposal on how to make it happen. I'd LOVE a PR for that functionality. I just haven't had time to address it.

phated:
I really don't think gulp will support this, so please make it optional @tkellen

tkellen:
It definitely would be. The more general case for this is just allowing liftoff to locate more than one configuration file. What you do with it is your business.

PR #1175, opened 2015-07-29
Adds a git.add step in order to work on windows and linux

klausbayrhammer:
the git.commit command as currently stated doesn't work on all environments. With windows and git version 1.9.5.msysgit.1 it works but with centos and git version 1.8.3.1 it doesn't. By adding the git.add() step it works in both environments


callumacrae:
The recipe was renamed: would you be able to update this PR to change the correct file?

alferov:
+1
Thanks, @klausbayrhammer, I had this issue last night when I tried this recipe.

callumacrae:
In the future, leave a comment when you push a new commit: Github sends notifications for comments but not updates!

PR #1185, opened 2015-08-03
Docs: Running Tasks In Series recipe 2nd example fix

michael-lehman:
Second example is supposed to show returning a stream instead of using a callback. Commit b65a70c3ec291cd89caa14a73796098e778fdb07 changed the 'clean' task to use a callback. The 'templates' and 'styles' do return streams, but 'clean' is the only task anything actually needs to wait for, making it the focus of the example.


callumacrae:
[del](https://www.npmjs.com/package/del) doesn't return a stream or promise

michael-lehman:
I apologize, made a bad assumption, but my original point remains. The second example is "supposed" to use a returned stream rather than a callback.

PR #1220, opened 2015-08-31
Update readme and recipes to reflect new Promise-oriented `del`

wbinnssmith:
Pending [sindresorhus/del#29](https://github.com/sindresorhus/del/pull/29) and [sindresorhus/vinyl-paths#3](https://github.com/sindresorhus/vinyl-paths/pull/3). Do not merge yet.

This updates the gulp readme to reflect an upcoming version of `del` that returns a promise rather than receives a callback. It ends up cleaning its usage through gulp quite nicely since tasks can return promises to complete them \o/

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,9 @@ var paths = {
 
 // Not all tasks need to use streams
 // A gulpfile is just another node program and you can use any package available on npm
-gulp.task('clean', function(cb) {
+gulp.task('clean', function() {
   // You can use multiple globbing patterns as you would with `gulp.src`
-  del(['build'], cb);
+  return del(['build']);
 });
 
 gulp.task('scripts', ['clean'], function() {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -38,9 +38,9 @@ var paths = {
 
 // Not all tasks need to use streams
 // A gulpfile is just another node program and you can use any package available on npm
-gulp.task('clean', function(cb) {
+gulp.task('clean', function() {
   // You can use multiple globbing patterns as you would with `gulp.src`
-  del(['build'], cb);
+  return del(['build']);
```
wbinnssmith: Also worth noting that `del` can accept a single string in these kinds of cases.

contra:
Will merge when released

README.md
sindresorhus: I think I explicitly added an array here because I noticed people kept calling `del` multiple times instead of using an array :facepalm:

sindresorhus:
Landed as both referenced modules have been updated.

sindresorhus:
687474703a2f2f74636c686f73742e636f6d2f36763641355a392e676966

PR #1221, opened 2015-08-31
Use the new del promise API

remcohaszing:
As of `del@2.0.0` it uses a promise instead of a callback function.


remcohaszing:
Duplicate of #1220

PR #1226, opened 2015-09-01
Add the chinese readme translation

awesomes-cn:
http://www.awesomes.cn is an open source for  github web repos, we offer translation for repo's readme and improve this to let our chinese users use the repo easily.In addition, we want more more developer join us to enrich the chinese readmes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 # gulp
 **The streaming build system**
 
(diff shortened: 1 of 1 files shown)


phated:
Simplified chinese docs already exist at https://github.com/gulpjs/gulp/tree/master/docs#gulp-documentation

PR #1259, opened 2015-09-16
Moving from vinyl-fs watch (gaze) to chokidar, fixes #1247

### LINKED ISSUE #1247: Watch method to proxy to chokidar
We need to update the `watch` method for gulp 4 to just use chokidar directly since vinyl-fs is removing the watch syntax in 2.0 (as per https://github.com/wearefractal/vinyl-fs/issues/92)

I know @es128 has mentioned he be able to help with this task.

ddprrt:
I gave integrating chokidar a shot. I hope you don't mind @es128. I see it as some exercise for me, so if you have any better solution, I don't mind having this PR going down ;-)

Naturally, I felt really uncomfortable changing tests ;-) But due to chokidar's different API, I had to. The changes were rather small, though. I also updated the recipes and docs.

Looking forward to your feedback

Fixes #1247

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -480,31 +480,42 @@ A single glob or array of globs that indicate which files to watch for changes.
 #### opts
 Type: `Object`
 
-Options, that are passed to [`gaze`][gaze].
+Options, that are passed to [`chokidar`][chokidar].
 
 #### fn
 Type: `Function`
 
 An [async](#async-support) function to run when a file changes.
 
-`gulp.watch` returns an `EventEmitter` object which emits `change` events with
-the [gaze] `event`:
+`gulp.watch` returns a wrapped [chokidar] FSWatcher object. If provided,
+the callback will be triggered upon any `add`, `change`, or `unlink` event.
+Listeners can also be set directly for any of [chokidar]'s events.
+
 ```js
 var watcher = gulp.watch('js/**/*.js', gulp.parallel('uglify', 'reload'));
-watcher.on('change', function(event) {
-  console.log('File ' + event.path + ' was ' + event.type);
+watcher.on('change', function(path, stats) {
+  console.log('File ' + path + ' was changed');
+  if (stats) {
+    console.log('changed size to ' + stats.size);
+  }
+});
+
+watcher.on('unlink', function(path) {
+  console.log('File ' + path + ' was removed');
 });
 ```
 
-##### event.type
+##### path
 Type: String
 
-The type of change that occurred, either "added", "changed" or "deleted".
+The relative path of the document.
 
-##### event.path
-Type: String
+##### stats
+Type: Object
 
-The path to the file that triggered the event.
+[File stats](http://nodejs.org/api/fs.html#fs_class_fs_stats) object when available.
+Setting the `alwaysStat` option to true will ensure that a file stat object will be
+available.
 
 ### gulp.tree(options)
 
@@ -708,7 +719,7 @@ module.exports = new MyCompanyTasksRegistry();
 ```
 
 [Function.name]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name
-[gaze]: https://github.com/shama/gaze
+[chokidar]: https://github.com/paulmillr/chokidar
 [glob-stream]: https://github.com/wearefractal/glob-stream
 [glob2base]: https://github.com/wearefractal/glob2base
 [gulp-if]: https://github.com/robrich/gulp-if
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@
 var util = require('util');
 var Undertaker = require('undertaker');
 var vfs = require('vinyl-fs');
+var chokidar = require('chokidar');
 
 function Gulp() {
   Undertaker.call(this);
@@ -30,7 +31,15 @@ Gulp.prototype.watch = function(glob, opt, task) {
     fn = this.parallel(task);
   }
 
-  return vfs.watch(glob, opt, fn);
+  var watcher = chokidar.watch(glob, opt);
+  if (fn) {
+    watcher
+      .on('change', fn)
+      .on('unlink', fn)
+      .on('add', fn);
+  }
+
+  return watcher;
 };
 
 // Let people use this class from our instance
--- a/test/watch.js
+++ b/test/watch.js
@@ -36,7 +36,7 @@ describe('gulp', function() {
       createTempFile(tempFile);
 
       var watcher = gulp.watch(tempFile, function(cb) {
-        watcher.end();
+        watcher.close();
         cb();
         done();
       });
@@ -44,13 +44,30 @@ describe('gulp', function() {
       updateTempFile(tempFile);
     });
 
+    it('should execute the gulp.parallel tasks', function(done) {
+      var tempFile = path.join(outpath, 'watch-func.txt');
+
+      createTempFile(tempFile);
+
+      gulp.task('test', function(cb) {
+        watcher.close();
+        cb();
+        done();
+      });
+
+      var watcher = gulp.watch(tempFile, gulp.parallel('test'));
+
+      updateTempFile(tempFile);
+    });
+
+
     it('should call the function when file changes: w/ options', function(done) {
       var tempFile = path.join(outpath, 'watch-func-options.txt');
 
       createTempFile(tempFile);
 
-      var watcher = gulp.watch(tempFile, {debounceDelay: 5}, function(cb) {
-        watcher.end();
+      var watcher = gulp.watch(tempFile, function(cb) {
+        watcher.close();
         cb();
         done();
       });
@@ -66,14 +83,11 @@ describe('gulp', function() {
 
       createTempFile(tempFile);
 
-      var watcher = gulp.watch(relFile, {debounceDelay: 5, cwd: cwd})
-        .on('change', function(evt) {
-          should.exist(evt);
-          should.exist(evt.path);
-          should.exist(evt.type);
-          evt.type.should.equal('changed');
-          evt.path.should.equal(path.resolve(tempFile));
-          watcher.end();
+      var watcher = gulp.watch(relFile, {cwd: cwd})
+        .on('change', function(filepath) {
+          should.exist(filepath);
+          path.resolve(cwd, filepath).should.equal(path.resolve(tempFile));
+          watcher.close();
           done();
         });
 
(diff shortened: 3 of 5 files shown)


test/watch.js
```diff
@@ -66,14 +83,11 @@ describe('gulp', function() {
...
-          should.exist(evt);
-          should.exist(evt.path);
-          should.exist(evt.type);
-          evt.type.should.equal('changed');
-          evt.path.should.equal(path.resolve(tempFile));
-          watcher.end();
+      var watcher = gulp.watch(relFile, {cwd: cwd})
+        .on('change', function(filepath) {
+          should.exist(filepath);
+          path.resolve(cwd, filepath).should.equal(path.resolve(tempFile));
```
ddprrt: This one was particular tricky, as chokidar spits out relative file paths instead of absolute ones.

index.js
```diff
@@ -30,7 +31,12 @@ Gulp.prototype.watch = function(glob, opt, task) {
     fn = this.parallel(task);
   }
 
-  return vfs.watch(glob, opt, fn);
+  var watcher = chokidar.watch(glob, opt);
+  if (fn) {
+    watcher.on('change', fn);
```
TrySound: I'd like to listen add, change and unlink events without additional code.

ddprrt: Yes, I was unsure of that one. `raw` might be a better choice

index.js
```diff
@@ -30,7 +31,12 @@ Gulp.prototype.watch = function(glob, opt, task) {
     fn = this.parallel(task);
   }
 
-  return vfs.watch(glob, opt, fn);
+  var watcher = chokidar.watch(glob, opt);
+  if (fn) {
+    watcher.on('all', fn);
```
ddprrt: @TrySound this should be much better

index.js
TrySound: @ddprrt I think better just set three listeners
```js
watcher
  .on('change', fn)
  .on('add', fn)
  .on('unlink', fn);
```

docs/API.md
```diff
@@ -480,31 +480,40 @@ A single glob or array of globs that indicate which files to watch for changes.
 #### opts
 Type: `Object`
 
-Options, that are passed to [`gaze`][gaze].
+Options, that are passed to [`chokidar`][chokidar].
```
callumacrae: You need to change the link at the bottom of the file

ddprrt: Whoops. Thanks! Added it.

docs/API.md
```diff
@@ -480,31 +480,40 @@ A single glob or array of globs that indicate which files to watch for changes.
...
+
 ```js
 var watcher = gulp.watch('js/**/*.js', gulp.parallel('uglify', 'reload'));
-watcher.on('change', function(event) {
-  console.log('File ' + event.path + ' was ' + event.type);
+watcher.on('change', function(path, stats) {
+  console.log('File ' + path + ' was changed');
+  if (stats) {
+    console.log('changed size to ' + stats.size);
+  }
```
callumacrae: It is possible to replicate the previous example in here?

es128:
> I hope you don't mind @es128

Not at all. Happy to have some help and to see this moving forward without me being the bottleneck.

test/watch.js
es128: Should be based on the input. If watching a relative path, events are emitted with matching relative paths. If you do `path.resolve(relFile)` at the top you should get absolute paths with the events.

index.js
es128: Definitely not `raw`. @TrySound's suggestion would be the correct approach.

index.js
es128: This will include `addDir` and `unlinkDir` events. Listening to the 3 events explicitly is a better match with what was there before.

es128: Take a look at [my original PR to glob-watcher](https://github.com/wearefractal/glob-watcher/pull/13/files) for a bit of reference. Although we are not going to want to keep re-mapping the divergent event names.

Unless someone can think of a reason why gulp users would want the dir events?

docs/API.md
```diff
@@ -480,31 +480,40 @@ A single glob or array of globs that indicate which files to watch for changes.
...
+Options, that are passed to [`chokidar`][chokidar].
 
 #### fn
 Type: `Function`
 
 An [async](#async-support) function to run when a file changes.
 
-`gulp.watch` returns an `EventEmitter` object which emits `change` events with
-the [gaze] `event`:
+`gulp.watch` returns an `FSWatcher` object that already listens to the `change`
```
es128: Suggested edit:

returns a wrapped [chokidar] `FSWatcher` object. If provided, the callback will be triggered upon any `add`, `change`, or `unlink` event. Listeners can also be set directly for any of [chokidar]'s events.

docs/API.md
```diff
@@ -480,31 +480,40 @@ A single glob or array of globs that indicate which files to watch for changes.
...
 
-The type of change that occurred, either "added", "changed" or "deleted".
+The relative path of the document.
 
-##### event.path
+##### stats
 Type: String
 
-The path to the file that triggered the event.
+Optional [file stats](http://nodejs.org/api/fs.html#fs_class_fs_stats) object (when available)
```
es128: A note could potentially be added about the `alwaysStat` option

docs/API.md
ddprrt: :+1: Changed it.

docs/API.md
ddprrt: :+1: Good call. I actually was expecting to have some more information :-)

index.js
ddprrt: I thought a little about it. The most basic pipeline is copying from A to B, and this includes files *and* folders, even if they're empty. So if you can pipe empty folders, you might also want to watch adding new directories in case you want to sync them on folder level.

However, this is also a rather rare use case. At the moment, the PR features @TrySound's suggestion.

docs/API.md
```diff
@@ -480,31 +480,42 @@ A single glob or array of globs that indicate which files to watch for changes.
...
-##### event.type
+##### path
 Type: String
 
-The type of change that occurred, either "added", "changed" or "deleted".
+The relative path of the document.
 
-##### event.path
+##### stats
 Type: String
```
es128: Type: Object

phated:
@es128 how are you feeling about this PR now?  Everything look similar to what your plans were?

@contra any feedback?

es128:
So long as you all are comfortable with the resulting API changes, then this implementation LGTM.

es128:
Oh, btw, I should be able to get chokidar 1.1.0 released next week. This will include fsevents 1.0.0, which now uses node-pre-gyp to make typical installs on Mac much faster and more stable.

docs/API.md
```diff
@@ -480,31 +480,42 @@ A single glob or array of globs that indicate which files to watch for changes.
...
+the callback will be triggered upon any `add`, `change`, or `unlink` event.
+Listeners can also be set directly for any of [chokidar]'s events.
+
 ```js
 var watcher = gulp.watch('js/**/*.js', gulp.parallel('uglify', 'reload'));
-watcher.on('change', function(event) {
-  console.log('File ' + event.path + ' was ' + event.type);
+watcher.on('change', function(path, stats) {
+  console.log('File ' + path + ' was changed');
+  if (stats) {
```
contra: can this be updated so stats is always present? don't like having this be optional

index.js
```diff
@@ -30,7 +31,15 @@ Gulp.prototype.watch = function(glob, opt, task) {
     fn = this.parallel(task);
   }
 
-  return vfs.watch(glob, opt, fn);
+  var watcher = chokidar.watch(glob, opt);
+  if (fn) {
```
contra: this fn is lacking docs - need more details about the opt and fn arguments

docs/API.md
```diff
@@ -480,31 +480,42 @@ A single glob or array of globs that indicate which files to watch for changes.
...
+Options, that are passed to [`chokidar`][chokidar].
 
 #### fn
 Type: `Function`
 
 An [async](#async-support) function to run when a file changes.
 
-`gulp.watch` returns an `EventEmitter` object which emits `change` events with
-the [gaze] `event`:
+`gulp.watch` returns a wrapped [chokidar] FSWatcher object. If provided,
```
contra: needs to pull in important (ie close) docs from the chokidar repo not just link to it - we have historically had issues just linking to other repos for docs - it makes discoverability hard

contra:
left comments

docs/API.md
es128: It would be a perf hit, and most users don't need/use it. Existing interface doesn't provide a `stat` object at all afaik, perhaps it just shouldn't be passed through here. 

Or perhaps *only* provide it if the user specified `alwaysStat`

contra: @es128 maybe just remove it from the example then, it's confusing

index.js
es128: `fn` is the user's optional callback, `opt` is the options object passed through to chokidar. Just like before. Both seem to me to be addressed in the README - you don't think it's adequate enough?

contra: @es128 the opt argument is 100% different now that the options are from chokidar and not gaze so those options need to be documented

es128: Ok cool so this relates to your other point about bringing in docs and not just linking to them. 

You called out documenting the `close()` method so far. Do you want all the options described here or just the most typically used ones (like `usePolling` for network shares)? I'd recommend the latter - with a link to the rest.

contra: @es128 most common w/ a link to the more advanced ones

docs/API.md
phated: :+1: on removing it from examples

docs/API.md
phated: :+1:

ddprrt:
Just want to let you know that I'll look into everything once I'm back from Italy ;-) (Wednesday the earliest)

es128:
@ddprrt cool. So if I have some time tonight or tomorrow I may also take a shot at addressing the feedback, starting with your commit as a basis. I believe it's only documentation changes left to do.

Enjoy the rest of your trip!

contra:
I'm going to be taking point on making sure this gets finished and merged FYI. Ping me when the changes have been pushed to the fork since GH doesn't notify on new commits.

es128:
You want a new PR with docs updates?

contra:
Merged this so we can get the functionality in, any docs stuff is a part of this https://github.com/gulpjs/gulp/issues/1281

phated:
We still need to explicitly bump chokidar to 1.1 in the deps so everyone gets the pre-gyp

es128:
Not strictly necessary, but sure.

contra:
@phated uses a caret range but I updated it anyways

PR #1264, opened 2015-09-17
Update Travis CI

ciceropablo:
See [`package.json`](https://github.com/gulpjs/gulp/blob/master/package.json#L63-L65).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 node_js:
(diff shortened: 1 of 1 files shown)


phated:
We are no longer testing on unstable versions of node

ciceropablo:
I can update engine to >= 0.10?

phated:
The engine field will be updated in gulp 4, but left for gulp 3.x

PR #1276, opened 2015-09-25
Added documentation about tasks which only have dependencies

mariuszskon:
It is possible to run tasks which only have dependencies, like so:
```js
gulp.task('build', ['array', 'of', 'task', 'names']);
```
So I added some documentation about it.


contra:
Should be clear that these will all run in parallel

mariuszskon:
How's this?

PR #1295, opened 2015-09-30
update vinyl-fs version

vineethawal:
Issue: Default value set for highWaterMark in **through2 npm** is 16(max buffer length), when default buffer length exceeds all files are not written by **vinyl-fs v0.3.x** dest().

This issue is resolved in **vinyl-fs v2.1.x**


phated:
Breaking version, you can use vinyl-fs directly in your gulpfile.  Just replace `gulp.src` with `vfs.src`, etc

PR #1299, opened 2015-10-01
documents glob negation

### LINKED ISSUE #1296: gulp.src ignore option does not match documentation
The [gulp API docs](https://github.com/gulpjs/gulp/blob/master/docs/API.md) indicate that gulp supports all options supported by node-glob:

> gulp adds some additional options in addition to the options supported by node-glob and glob-stream:

[Node-glob](https://github.com/isaacs/node-glob#options) says it supports an `ignore` option.

> ignore Add a pattern or an array of patterns to exclude matches.

Furthermore, node-glob says to use `ignore` and not to use negation:

> The intent for negation would be for a pattern starting with `!` to match everything that doesn't match the supplied pattern. However, the implementation is weird. It is better to use the `ignore` option to set a pattern or set of patterns to exclude from matches.

However, as indicated in issue #1176, `ignore` does not work in gulp. Something is broken: the docs, the implementation, or both. Step one is to determine the desired state. Step two is to update the docs and/or code. What should the behavior be?

breyed:
Fixes #1296

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -20,33 +20,43 @@ gulp.src('client/templates/*.jade')
   .pipe(gulp.dest('build/minified_templates'));
 ```
 
-`glob` refers to [node-glob syntax](https://github.com/isaacs/node-glob) or it can be a direct file path.
-
 #### globs
 Type: `String` or `Array`
 
-Glob or array of globs to read.
+Glob or array of globs to read. Globs use [node-glob syntax](https://github.com/isaacs/node-glob) except that negation is fully supported.
+
+A glob that begins with `!` excludes matching files from the glob results up to that point. For example, consider this directory structure:
+
+    client/
+      a.js
+      bob.js
+      bad.js
+
+The following expression matches `a.js` and `bad.js`:
+
+    gulp.src(['client/*.js', '!client/b*.js', 'client/bad.js'])
+    
 
 #### options
 Type: `Object`
 
 Options to pass to [node-glob] through [glob-stream].
 
-gulp adds some additional options in addition to the [options supported by node-glob][node-glob documentation] and [glob-stream]:
+Gulp supports all [options supported by node-glob][node-glob documentation] and [glob-stream] except `ignore` and adds the following options.
 
-#### options.buffer
+##### options.buffer
 Type: `Boolean`
 Default: `true`
 
 Setting this to `false` will return `file.contents` as a stream and not buffer files. This is useful when working with large files. **Note:** Plugins might not implement support for streams.
 
-#### options.read
+##### options.read
 Type: `Boolean`
 Default: `true`
 
 Setting this to `false` will return `file.contents` as null and not read the file at all.
 
-#### options.base
+##### options.base
 Type: `String`
 Default: everything before a glob starts (see [glob2base])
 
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -20,33 +20,43 @@ gulp.src('client/templates/*.jade')
...
+    gulp.src(['client/*.js', '!client/b*.js', 'client/bad.js'])
+    
 
 #### options
 Type: `Object`
 
 Options to pass to [node-glob] through [glob-stream].
 
-gulp adds some additional options in addition to the [options supported by node-glob][node-glob documentation] and [glob-stream]:
+Gulp supports all [options supported by node-glob][node-glob documentation] and [glob-stream] except `ignore` and adds the following options.
```
contra: gulp is lowercase in the docs

contra:
Looks good!

docs/API.md
breyed: I wasn't sure if the lowercase *g* was a typo. So I headed over to gulpjs.com. The big cursive *G* on the landing page led me to believe that Gulp sees itself as a typical proper noun. It would be nice if the docs followed this, as the capital letter makes the sentences easier to parse.

contra: @breyed There is a page in the docs about this (in the FAQ) - gulp is always lowercase, the only exception is the logo and that's because the person who made the logo did it for aesthetic reasons

docs/API.md
```diff
@@ -20,33 +20,43 @@ gulp.src('client/templates/*.jade')
   .pipe(gulp.dest('build/minified_templates'));
 ```
 
-`glob` refers to [node-glob syntax](https://github.com/isaacs/node-glob) or it can be a direct file path.
-
 #### globs
 Type: `String` or `Array`
 
-Glob or array of globs to read.
+Glob or array of globs to read. Globs use [node-glob syntax](https://github.com/isaacs/node-glob) except that negation is fully supported.
```
callumacrae: Is it worth moving the url to the bottom of the file? It's like that for the rest of the links in this file :)

docs/API.md
```diff
@@ -20,33 +20,43 @@ gulp.src('client/templates/*.jade')
...
+    gulp.src(['client/*.js', '!client/b*.js', 'client/bad.js'])
+    
 
 #### options
 Type: `Object`
 
 Options to pass to [node-glob] through [glob-stream].
 
-gulp adds some additional options in addition to the [options supported by node-glob][node-glob documentation] and [glob-stream]:
+gulp supports all [options supported by node-glob][node-glob documentation] and [glob-stream] except `ignore` and adds the following options.
```
callumacrae: Could you put the colon back? Not important, just a bit clearer.

breyed: A colon precedes a block or list within a section. Text preceding subsections does not use a colon, since the scope of what the colon applies to would not be clear.

Examples from HTML5 spec: [List](http://www.w3.org/TR/html5/introduction.html#extensibility), [Subsections (microsyntaxes)](http://www.w3.org/TR/html5/infrastructure.html#common-microsyntaxes), [Subsections (collections)](http://www.w3.org/TR/html5/infrastructure.html#collections), [List and subsection (kind of content)](http://www.w3.org/TR/html5/dom.html#kinds-of-content)

callumacrae: It's a list of options and is perfectly clear ಠ_ಠ

breyed: It's a list of options where the items have multiple paragraphs and a code block - enough complexity that they are laid out as subsections. It's a fine point, and a debatable one, but my opinion is that a colon isn't the right grammar in this case.

PR #1312, opened 2015-10-07
User note

SibuStephen:
There may be chances in some cases, where one may need to use sudo (Not a necessity) or in case of windows with administrative Privilege.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -5,6 +5,7 @@
 ```sh
 $ npm install --global gulp
 ```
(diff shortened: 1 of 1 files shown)


contra:
The npm docs specifically say not to install as root though, so this shouldn't be needed. If you have to use sudo you installed npm incorrectly.

phated:
Agree with @contra - it is actually a very bad sign if you require root.

SibuStephen:
@contra  @phated  Agreed with it.

PR #1374, opened 2015-11-06
Changed increment operators in tests to shorthand addition.

sohamkamani:
Increment operators are [deprecated](https://jslinterrors.com/unexpected-plus-plus) in several javascript linting tools. The increment operators in the test cases are simply changed to shorthand addition (+= 1)


phated:
Thanks, but I think we will keep it the current way. The feature is in the language and is not deprecated.

PR #1380, opened 2015-11-09
Add browserify (multiple destination) example

kt3k:
[Browserify+Globs](https://github.com/gulpjs/gulp/blob/master/docs/recipes/browserify-with-globs.md) recipe bundles everything into one bundle, but doesn't show how to bundle each file into a separate bundle.

I think this example is helpful for people who want to create separate bundles from different entry files.

[This blog post](https://wehavefaces.net/gulp-browserify-the-gulp-y-way-bb359b3f9623) covers almost the same thing, but because of the change of browserify API now the examples in the above post doesn't work. (see https://github.com/substack/node-browserify/issues/1217 )

I believe it's helpful for beginners to have this example in the official recipes.

Thanks,

### DIFF
--- a/docs/recipes/browserify-multiple-destination.md
+++ b/docs/recipes/browserify-multiple-destination.md
@@ -16,7 +16,7 @@ var uglify = require('gulp-uglify');
 
 gulp.task('js', function () {
 
-  return gulp.src('src/**/*.js', {base: 'src'})
+  return gulp.src('src/**/*.js', {read: false}) // no need of reading file because browserify does.
 
     // transform file objects using gulp-tap plugin
     .pipe(tap(function (file) {
--- a/docs/recipes/browserify-multiple-destination.md
+++ b/docs/recipes/browserify-multiple-destination.md
@@ -7,16 +7,19 @@ The below `js` task bundles all the `.js` files under `src/` as entry points and
 
 ```js
 var gulp = require('gulp');
-var plugins = require('gulp-load-plugins')();
 var browserify = require('browserify');
 var gutil = require('gulp-util');
+var tap = require('gulp-tap');
+var buffer = require('gulp-buffer');
+var sourcemaps = require('gulp-sourcemaps');
+var uglify = require('gulp-uglify');
 
 gulp.task('js', function () {
 
   return gulp.src('src/**/*.js', {base: 'src'})
 
-    // transform file objects using gulp-map plugin
-    .pipe(plugins.tap(function (file) {
+    // transform file objects using gulp-tap plugin
+    .pipe(tap(function (file) {
 
       gutil.log('bundling ' + file.path);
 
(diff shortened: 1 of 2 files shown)


docs/recipes/browserify-multiple-destination.md
```diff
@@ -0,0 +1,43 @@
...
+var gulp = require('gulp');
+var plugins = require('gulp-load-plugins')();
+var browserify = require('browserify');
+var gutil = require('gulp-util');
+
+gulp.task('js', function () {
+
+  return gulp.src('src/**/*.js', {base: 'src'})
+
+    // transform file objects using gulp-map plugin
```
noamokman: you mean `gulp-tap` right?

phated:
This seems to be a fairly clean example.  Going to think on it a bit.

docs/recipes/browserify-multiple-destination.md
```diff
@@ -0,0 +1,43 @@
+# Browserify + Globs (multiple destination)
+
+This example shows how to set up a task of bundling multiple entry points into multiple destinations using browserify.
+
+The below `js` task bundles all the `.js` files under `src/` as entry points and writes the results under `dest/`.
+
+
+```js
+var gulp = require('gulp');
+var plugins = require('gulp-load-plugins')();
```
contra: probably shouldnt use gulp-load-plugins in the recipes, just to keep them agnostic

kt3k: OK. I'll change them.

docs/recipes/browserify-multiple-destination.md
kt3k: Yes! Thanks!

It was `gulp-map` at first but I changed it later.

kt3k:
Does this need any further fix or improvement?

phated:
@kt3k probably add `read: false` to the options so the files aren't read during `gulp.src` globbing

kt3k:
Ah, that makes sense. This reads the same file twice. I'll change it. thanks.

kt3k:
Updated the change. Added some comment about `read: false` option.
Removed `base: 'src'` option because it seemed unnecessary.

kt3k:
@phated 
I updated the example. Added some comment about `read: false` option.
Removed `base: 'src'` option because it seemed unnecessary.

Kiura:
@kt3k, Hi! sorry to bother you like that but any chance you (or anyone else) could help me with setup for gulp?
basically what I am trying to do is babelify and browserify files;
The structure of my project: 
```javascript
--js
   |-- index.js
   |-- login.js
--final
   |-- index-bundle.js // this is what i need
   |-- login-bundle.js // this is what i need
```

I can accomplish that: (but it probably will be slow)
method 1:
```javascript
gulp.task('es6',function (done) {
	glob('./public/js/*.js', function(err, files) {
        if(err) done(err);
        var tasks = files.map(function(entry) {
            return browserify({ entries: [entry] })
				.transform(babelify)
                .bundle()
                .pipe(source(entry))
                .pipe(rename(function (path) {
					path.dirname = "final";
				}))
                .pipe(gulp.dest('./public/js/'));
            });
        es.merge(tasks).on('end', done);
    })
});
```
A better approach would be to use watchify (or maybe something else i am not aware of?)
method 2:
```javascript
var files = glob.sync('./public/js/*.js');
var customOpts = {
	entries: files,
	debug: true
};
var opts = assign({}, watchify.args, customOpts);
var b = watchify(browserify(opts)); 
b.transform(babelify);
b.on('update', bundle); // on any dep update, runs the bundler
b.on('log', gutil.log); // output build logs to terminal	

function bundle() {
  return b.bundle()
    .pipe(source('bundle.js'))
    .pipe(rename(function (path) {
		path.dirname = "final";
	}))
    .pipe(gulp.dest('./public/js'));
}
```
This approach is faster on rebuilds, however id bundles everything into one file.
The best solution would be to change method 2 so it would output multiple files, but any other solutions highly appreciated.

PR #1218, opened 2015-08-28
Upgrade semver because it's insecure

fscherwi:
https://nodesecurity.io/advisories/semver_redos

PR #1387, opened 2015-11-12
upgrade semver to newer version

fscherwi:
See #1218!


fscherwi:
#1218

phated:
There's no reason for this. The arrow will cause that version to be picked up.

PR #1388, opened 2015-11-12
Improve task not found messaging

soulcutter:
This advice seems more-likely to be helpful when mistakenly referencing unknown tasks.


contra:
Can you make this change to the 4.0 branch instead of master?

soulcutter:
Sure

PR #1395, opened 2015-11-16
Update browserify-with-globs.md

fchastanet:
when browserify throw this error
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: path must be a string
    at /.../node_modules/browserify/node_modules/resolve/lib/async.js:16:16
    at doNTCallback0 (node.js:428:9)
    at process._tickCallback (node.js:357:13)

we need some debugging information : added on('error'...) on the bundle


phated:
Recipes are supposed to be simplified to get you started and don't include common error handling logic.

Issue #1401, opened 2015-11-22
How to use gulpfile.babel.js is undocumented


shadowmint:
I realize this isn't a commonly used feature, but this functionality seems to have no documentation for it *anywhere* outside of this issue tracker.

Upgrading packages just resulted in complete breakage of my gulp workflow due to https://github.com/gulpjs/gulp/issues/1364

Please consider adding an article to https://github.com/gulpjs/gulp/tree/master/docs/recipes that documents that need for a `.babelrc` file, and broadly speaking, how to use es6 gulpfiles in general.

callumacrae:
It's documented here: https://github.com/gulpjs/gulp/blob/master/docs/CLI.md#compilers

But yeah, that could be clearer. I'll look to add a recipe (or if you want to contribute one, feel free!).

It's worth noting that there's not much a documentation improvement with gulp could have done about the Babel 6 update.


I've also written an article about it on my website: http://macr.ae/article/gulp-and-babel.html

phated:
It's not our responsibility to document other libraries you might be including in your project.

shadowmint:
@phated This is a core feature of gulp. Either don't support it, or document it.

PR #1403, opened 2015-11-22
Add recipe for Using ES6 with Gulp.

carloscuesta:
Add recipe for Using ES6 with Gulp. See #1401


phated:
This doesn't work with old babel and we aren't responsible for documenting other libraries that someone might include in their projects.  I would be more inclined to include a link to @callumacrae's blog post about this.

PR #1424, opened 2015-12-02
Surface 'on' listener functionality

AndrewSouthpaw:
It is not apparent from the docs that src can be listened for when it finishes, which is immensely useful if performing a pipe after an asynchronous result. I've added an example to the documentation to surface this functionality, since I lost several hours puzzling over how to accomplish this task.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -163,14 +163,12 @@ gulp.task('jekyll', function(cb) {
     cb(); // finished task
   });
 });
-```
 
-##### Use a callback after a stream ends
-
-```javascript
+// use an async result in a pipe
 gulp.task('somename', function(cb) {
-  someAsyncAction(function(res) {
-    var stream = gulp.src('client/**/*.js')
+  getFilesAsync(function(err, res) {
+    if (err) return cb(err);
+    var stream = gulp.src(res)
       .pipe(minify())
       .pipe(gulp.dest('build'))
       .on('end', cb);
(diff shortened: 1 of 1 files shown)


phated:
This is underlying streams functionality.  Instead of just an example, that should be explained and streams documentation should be linked from https://nodejs.org/api/

contra:
I think this is fair to add to our documentation, we could argue everything is obvious but the whole point is to help the user write better gulpfiles

phated:
@contra I didn't say we shouldn't add it, just that it needs more than an example

ilanbiala:
@phated what else needs to be added?

contra:
@AndrewSouthpaw Can you add this to the 4.0 docs (4.0 branch) instead of the master branch?

phated:
I'm fine with this getting into the master branch but I still think we need more of just an example.  At very minimum, a sentence or 2 explaining what is happening and a link to the node docs for events that streams emit.

phated:
Still waiting on a sentence or 2 explaining what is happening with a link to the node docs for events that streams emit.

ilanbiala:
@phated https://github.com/AndrewSouthpaw seems to be inactive based on his last commit being this one, so maybe just do it for him?

phated:
@ilanbiala feel free to submit another PR and I will close this one

AndrewSouthpaw:
Sorry, I'm just bad at checking my GitHub notifications. ;-) Thanks for merging in and taking care of the followup on my behalf.

Issue #1396, opened 2015-11-17
Documenting gulp.task arguments


jacopotarantino:
Could someone please explain the callback fed to `gulp.task`? I'm not sure if it's supposed to take arguments, what it's supposed to do, or if gulp depends on my plugin/function listening for this callback function to invoke a/sync responses. I was looking at stackoverflow as well as https://github.com/gulpjs/gulp/blob/master/docs/API.md#gulptaskname-deps-fn and couldn't find a sufficient explanation of this information.

kevdez:
The callback can be whatever you want it to be, whether you want to accept arguments or not. Usually the callback consists of what you want that gulp task to do, after it's done with the array of dependency tasks (if any).

I noticed that the documentation for gulp.task() is a little off from the rest of the API documentation. I'll fix that.

jacopotarantino:
@kevdez thanks! Undocumented function arguments can be really confusing.

phated:
Closed by #1429

PR #1429, opened 2015-12-04
Fixed docs on gulp.task's optional arguments & provided a better example

kevdez:
I noticed that gulp.task docs don't have the task arguments' `Type: ` sections, so I added that.

I also put in a better example for a `gulp.task(name, function ( ) { } )` use case. The previous single line example was a little too simplified tbh.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -108,7 +108,7 @@ Default: `0777`
 
 Octal permission string specifying mode for any folders that need to be created for output folder.
 
-### gulp.task(name[, deps], fn)
+### gulp.task(name [, deps, fn])
 
 Define a task using [Orchestrator].
 
@@ -119,6 +119,7 @@ gulp.task('somename', function() {
 ```
 
 #### name
+Type: `String`
 
 The name of the task. Tasks that you want to run from the command line should not have spaces in them.
 
@@ -135,7 +136,7 @@ gulp.task('mytask', ['array', 'of', 'task', 'names'], function() {
 
 **Note:** Are your tasks running before the dependencies are complete?  Make sure your dependency tasks are correctly using the async run hints: take in a callback or return a promise or event stream.
 
-You can also omit the function if you only want to run the dependency tasks:
+You can also omit the function if you only want to run a bundle of dependency tasks:
 
 ```js
 gulp.task('build', ['array', 'of', 'task', 'names']);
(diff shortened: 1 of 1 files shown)


kevdez:
I would also close #1396 for this, since it addresses that issue.

docs/API.md
```diff
@@ -144,8 +145,19 @@ gulp.task('build', ['array', 'of', 'task', 'names']);
...
 #### fn
+Type: `Function`
+
+The function that performs the task's main operations. Generally this takes the form of:
 
-The function that performs the task's operations. Generally this takes the form of `gulp.src().pipe(someplugin())`.
+```js
+gulp.task('buildStuff', function(someArguments) {
+  // Do something that "builds stuff"
+  gulp.src(/*some source path*/)
```
phated: always return your stream

docs/API.md
```diff
@@ -144,8 +145,19 @@ gulp.task('build', ['array', 'of', 'task', 'names']);
 **Note:** The tasks will run in parallel (all at once), so don't assume that the tasks will start/finish in order.
 
 #### fn
+Type: `Function`
+
+The function that performs the task's main operations. Generally this takes the form of:
 
-The function that performs the task's operations. Generally this takes the form of `gulp.src().pipe(someplugin())`.
+```js
+gulp.task('buildStuff', function(someArguments) {
```
phated: no arguments are passed here other than callback and that's only if you specify the `cb` argument

phated: no arguments are passed here other than callback and that's only if you specify the `cb` argument

docs/API.md
```diff
@@ -108,7 +108,7 @@ Default: `0777`
 
 Octal permission string specifying mode for any folders that need to be created for output folder.
 
-### gulp.task(name[, deps], fn)
+### gulp.task(name [, deps] | [, deps[, fn]] | [, fn])
```
phated: I'm confused by the function signature you changed this to I don't think we support an array of deps as the third argument

kevdez: The bracket notation means it's optional, nothing to mean that it's an array. Whatever `deps` is, is already explained later.
`gulp.task(name [, deps] | [, deps[, fn]] | [, fn])`
translates to:
`gulp.task(name, deps)`
or
`gulp.task(name, deps, fn)`
or
`gulp.task(name, fn)`

docs/API.md
kevdez: The original documentation also has a `fn` argument, no `cb` anywhere. I also thought this was a little inconsistent with the rest of the docs, but I kept it as `fn`.

docs/API.md
```diff
@@ -108,7 +108,7 @@ Default: `0777`
 
 Octal permission string specifying mode for any folders that need to be created for output folder.
 
-### gulp.task(name[, deps], fn)
+### gulp.task(name [, deps] | [, deps[, cb]] | [, cb])
```
phated: I don't think we need the `|` and each signature as separate because the `[ ]` syntax is commonplace for optional

kevdez: Okay, I can take those off. 

The `fn` is not in `[ ]` and yet later in the documentation it says it's optional, for example in `gulp.task('build', ['list','of','deps'])` so shouldn't `fn` be optional as well?

phated: yeah, `fn` is optional as well

kevdez: Yeah, so which is better `gulp.task(name [, deps] [, fn])` or `gulp.task(name [, deps, fn])`? But then that means `gulp.task(name)` is valid, which isn't very useful. Also, to keep consistency, should it be `fn` or `cb` like in `gulp.watch`?

phated:
Thanks!

PR #1456, opened 2015-12-26
Bump interpret to 1.0.0

duereg:
This will allow gulp to work with the new `babel-require` library for `*.babel.js` files


phated:
This should be good.  I believe we handled backward-compat correctly when going 1.0

teohhanhui:
It is `babel-register`, not `babel-require`.

PR #1462, opened 2015-12-28
Another gulp4 with full back compatibility

zensh:
Hi, all gulp man, this is another version of gulp v4.
Here is document: https://github.com/zensh/gulp/blob/thunk-gulp/docs/API.md

TODO:
1. more test;
2. recursion detect.


callumacrae:
A PR this big is not very helpful and impossible to review. Also, a lot of changes you've made seem to changing the code standards, which afaik we don't want to do.

If you have any useful changes, please break them into individual pull requests!

Issue #1281, opened 2015-09-25
pull in external docs


contra:
right now all of the docs are scattered amongst sub-repos, and our main repo just links to them. this is really really shitty and makes it difficult to know where to file issues and difficult to find documentation.

repos that need to be pulled in:
- [x] vinyl-fs (for .src, .dest, and .symlink)
  - [x] vinyl (for what files coming out of .src and .dest look like)
  - [x] glob-stream (for .src)
- [x] chokidar (for .watch)
- [x] undertaker

es128:
Ok. Let me see if I can help with the chokidar stuff tonight.

phated:
Looking for some more people to contribute on the remaining items in this list.

troch:
I thought I had commented on this, must have forgotten to click "Comment"... Looking at API.md on 4.0 branch, undertaker is referenced and its methods (task, lastRun, series, parallel) seem to have up-to-date documentation, but they are scattered. Should those API functions be grouped together, API.md split into smaller files, have a main file describing Gulp's main dependencies making its API...?

Romanx:
I'll take a look at those that aren't pulled in and make a PR

phated:
Ref #1446

phated:
The projects listed (and even more) have been incorporated into our API Docs rewrite so I'm going to close this.

Note: We brought them all in by hand so they'll have to be kept up-to-date manually.  That being said, most are controlled by us or don't change very often.

PR #1465, opened 2015-12-30
Add glob-stream, node-glob docs to API docs

doug-wade:
[1281](https://github.com/gulpjs/gulp/issues/1281)

I erred on the side of including too many options instead of too few, to make reviewing easier -- all of them are included here that are supported, so you shouldn't have to open the docs -- so likely we'll want to prune them.  I'd say all the cache reuse stuff (symlinks, statCache, and buffer), and all the "don't include this glob feature" options (who doesn't like more features 😀) at least should be removed.

I didn't include any of the references to node-glob, since I don't think we want to leak implementation details from _our_ dependencies, since they can be changed in a semver patch release, and all our dependencies are caret ranges, but it does add a lot of weight to the gulp.src docs; we may want to remove a lot of the glob primer.

Sorry the diffs a little noisy -- alphabetizing the options simultaneously made it easier to write, but harder to review.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -28,75 +28,113 @@ gulp.src('client/templates/*.jade')
 #### globs
 Type: `String` or `Array`
 
-Glob or array of globs to read.
+Glob or array of globs to read. Globs use the following syntax
 
-Note that globs are evaluated in order, which means this is possible:
+"Globs" are the patterns you type when you do stuff like `ls *.js` on
+the command line, or put `build/*` in a `.gitignore` file.
+
+Before parsing the path part patterns, braced sections are expanded
+into a set.  Braced sections start with `{` and end with `}`, with any
+number of comma-delimited sections within.  Braced sections may contain
+slash characters, so `a{/b/c,bcd}` would expand into `a/b/c` and `abcd`.
+
+The following characters have special magic meaning when used in a
+path portion:
+
+* `*` Matches 0 or more characters in a single path portion
+* `?` Matches 1 character
+* `[...]` Matches a range of characters, similar to a RegExp range.
+  If the first character of the range is `!` or `^` then it matches
+  any character not in the range.
+* `!(pattern|pattern|pattern)` Matches anything that does not match
+  any of the patterns provided.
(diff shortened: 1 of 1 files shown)


doug-wade:
I poked around the failing build on 0.10.x -- it's failing with linting errors due to conflicts in how strict mode is handled, but I didn't touch bin/gulp?  Is there some way to kick of another build, to check if this is a transient failure?  Or did I actually break something?

stevemao:
This should be done automatically by a build script.

ilanbiala:
Any progress on this?

contra:
Yeah, we can ignore the tests - something is whack with the linting lately. LGTM. @phated?

contra:
I'm starting to think we should break the docs apart and use something like gitbook to generate the webpage

phated:
@contra are you not interested in using readme.io anymore?

contra:
@phated I think not, having to manually write the docs + them not being editable by the community is a dealbreaker. We can make a custom skin + plugins for gitbook just as easily.

contra:
Here's an example gitbook auto-create setup I made:

- System: https://github.com/shastajs/palisade/blob/master/package.json#L40-L43
- Output: http://shasta.tools/palisade/

On any version bump it will create new docs and publish them to gh-pages

doug-wade:
@contra should I close the pr since we're going to migrate to gitbook, which afaict would move the docs into a separate repo altogether and make this obsolete?  Or would it make sense to merge these changes and consider a different doc solution separately?  I _personally_ prefer to have my docs as close as possible to the code that they document -- ideally as doc strings at the method level, but at least in the same file and definitely in the same repo -- since it makes it easier to remember to update the docs when you're changing the code, but kinda feel like that discussion belongs in an issue rather than this pr (and also that I don't have a lot of the necessary context to contribute helpfully :stuck_out_tongue_winking_eye:).

phated:
I'm going to close this, as we should implement an automated way using a plugin in gitbook that fetches a markdown file from github and clips out a section based on a header.

Issue #1487, opened 2016-01-14
Gulp 4: gulp.parallel will stop gulp immediately when one task fails


per2plex:
In the [docs](https://github.com/gulpjs/gulp/blob/4.0/docs/API.md#gulpparalleltasks) it is stated that gulp.parallel will still execute all tasks if an error occurs:
> When the returned function is executed, the tasks or functions will be executed in parallel, all being executed at the same time. **If an error occurs, all execution will complete.**

This currently doesn't seem to be the case as the following minimal code example will demonstrate:

```javascript
var gulp = require('gulp')

gulp._settle = false

gulp.task('error', function(done) {
  setTimeout(function() {
    done(new Error('Somethings wrong!'))
  }, 200)
})

gulp.task('willnotfinish', function(done) {
  setTimeout(function() {
    console.log('I will never finish :(')
    done()
  }, 2000)
})

gulp.task('watch', function() {
  gulp.watch('watchme.js', gulp.parallel('default'))
})

gulp.task('default', gulp.parallel(
  'error',
  'willnotfinish'
))
```
This will stop running gulp after ~200ms with the following output and the task 'willnotfinish' will never finish executing:
```
Using gulpfile ~/Desktop/development/gulp-parallel-error/gulpfile.js
Starting 'default'...
Starting 'error'...
Starting 'willnotfinish'...
'error' errored after 204 ms
Error: Somethings wrong!
  at null._onTimeout (/Users/perplex/Desktop/development/gulp-parallel-error/gulpfile.js:5:10)
  at Timer.listOnTimeout (timers.js:92:15)
'default' errored after 211 ms
The following tasks did not complete: willnotfinish
Did you forget to signal async completion?
```

EDIT: Include watch task in code sample

per2plex:
This actually seems to be solved by setting the `UNDERTAKER_SETTLE` environment variable to `true`, is there a reason `bach.settleParallel` is not used by default?

per2plex:
Works with #1488 and `_settle` set to `true`. Also seems to work fine with my project which incorporates several gulp plugins, so i'm still unsure about the side effects of `_settle = true` if there are any.

I updated the code sample above to include a watch task.

phated:
[...]

PR #1488, opened 2016-01-14
Wrap chokadir callbacks in an additional function to solve some under…

per2plex:
…lying problems with undertaker and bach in combination with undertaker's _settle flag.

Chokadir passes the path of the added/removed/changed file to his event callbacks which will end up being the `done` parameter inside bach's `buildParallel` `parallel` function, this doesn't seem to be noticeable without the `_settle` flag set to `true` but could lead to some future hard to trace down bugs.

Solves #1487 partially.

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,9 +47,9 @@ Gulp.prototype.watch = function(glob, opt, task) {
   var watcher = chokidar.watch(glob, opt);
   if (fn) {
     watcher
-      .on('change', fn)
-      .on('unlink', fn)
-      .on('add', fn);
+      .on('change', function() { fn(); })
+      .on('unlink', function() { fn(); })
+      .on('add', function() { fn(); });
   }
 
   return watcher;


contra:
LGTM

phated:
I think this bug is actually in undertaker and we shouldn't be adding extra closures before looking deeper.

per2plex:
The reason it works when [`buildParallel`](https://github.com/gulpjs/bach/blob/master/lib/parallel.js) instead of [`buildSettleParallel`](https://github.com/gulpjs/bach/blob/master/lib/settleParallel.js) is used, is that [`map`](https://github.com/gulpjs/now-and-later/blob/master/lib/map.js) in the `now-and-later` module [checks](https://github.com/gulpjs/now-and-later/blob/master/lib/map.js#L14-L17) if the `done` parameter actually is a function.

`buildParallel` passes the `done` parameter [directly](https://github.com/gulpjs/bach/blob/master/lib/parallel.js#L19) to `nowAndLater.map`.
`buildSettleParallel` instead [passes](https://github.com/gulpjs/bach/blob/master/lib/settleParallel.js#L19) the `done` first to a helper function `helpers.onSettled` which [doesn't check](https://github.com/gulpjs/bach/blob/master/lib/helpers.js#L14) if `done` actually is a function, which in this case it won't be because it will be set to the path of the changed file from chokidar, so a check if `done` is actually a function in there would also solve the problem.

phated:
@per2plex yep, that should be the solution we use instead of adding extra closures.  Thanks for digging into this!

per2plex:
I'll close this pull request in that case and open another one with the proposed fix on bach instead.

phated:
The fix for this was published as bach 0.4.2 - Thanks again.

PR #1494, opened 2016-01-22
Updating docs to reflect 4.0 syntax

beenanner:


phated:
`default` is a reserved word in JS.  You can't name a function that.  Also, you don't return the `done` callback, just streams/promises/etc.

beenanner:
Sorry @phated bad form on my part. I opened a new pull-request with the corrections.

PR #1501, opened 2016-01-26
Make watch() work without opts or callback

jmm:
Except for the method signature in the end user docs, every indication is that `watch()` is supposed to work without a callback. But as a result of #1283, if you pass neither opts nor a callback it throws. This fixes that and updates the documentation of the method signature, and the prose about how callback is invoked.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -481,7 +481,7 @@ Type: `Array`, `String` or `Function`
 A task name, a function or an array of either.
 
 
-### gulp.watch(glob[, opts], fn)
+### gulp.watch(glob[, opts][, fn])
 
 Watch files and do something when a file changes.
 File watching is provided by the [`chokidar`][chokidar] module.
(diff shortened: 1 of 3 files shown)


docs/API.md
```diff
@@ -481,7 +481,7 @@ Type: `Array`, `String` or `Function`
 A task name, a function or an array of either.
 
 
-### gulp.watch(glob[, opts], fn)
+### gulp.watch(glob[, opts][, fn])
```
contra: should be `glob[, opts, fn]` right?

contra:
LGTM

docs/API.md
callumacrae: No, because that implies that if you want to specify `fn`, you have to specify `opts`, and that's not true. They're independently optional and so `glob[, opts][, fn]` is correct :)

phated: I agree with @callumacrae

PR #1506, opened 2016-01-29
Add exports as tasks docs

### LINKED ISSUE #1445: Document exports as task
We should add some documentation for this.  Not sure if it should be here, or gulp-cli.  Maybe it would be a good recipe.

Ref https://github.com/gulpjs/gulp/issues/1305

cc @thejameskyle

thejameskyle:
Closes #1445


phated:
Thanks @thejameskyle!

PR #1510, opened 2016-01-31
suggest gulp-cli in the getting started

### LINKED ISSUE #1451: Should we suggest gulp-cli for 3.x?
Currently, gulp ships with the CLI built in, but we split it out to allow for different versions of gulp to be handled.  It is going to be suggested in 4.x to install gulp-cli globally instead of gulp.  However, should we update the docs in master (for 3.x) to suggest that also?

phated:
Closes #1451 

@contra How does this look?

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -2,8 +2,11 @@
 
 #### 1. Install gulp globally:
 
+__If you have previously installed a version of gulp globally, please run `npm rm --global gulp`
+to make sure your old version doesn't collide with gulp-cli.__
+
 ```sh
-$ npm install --global gulp
+$ npm install --global gulp-cli
 ```
 
 #### 2. Install gulp in your project devDependencies:


contra:
@phated is the collision an issue?

phated:
In old versions of npm (ones that ship with many versions of node), if you did:
1. `npm install -g gulp`
2. `npm install -g gulp-cli`
3. `npm rm -g gulp`
It would remove the gulp bin because it knew that gulp could create it but it didn't check if the bin was owned by gulp.  We had this fixed as part of our issues in a 3.x release but I don't know if it ever got backported to 2.x and many people, myself included, still use 2.x for speed, etc.

contra:
@phated Mind testing it on 2.x and reporting if it's an issue?

phated:
I'd prefer not to, I have gulp-cli linked globally to work on it and the switching between linked and installed is a nightmare due to manpage generation.  Even if this was fixed in 2.x, 1.4.x is still out in the wild and shipping with versions of node (0.10.x)

contra:
@phated okay.

PR #1513, opened 2016-02-03
make comments easier to comprehend

LoganArnett:
Just felt that this may allow first time/early gulp users to comprehend what each line is doing a little more than having the comments as a block above. Having the line by line may help connect the dots.

### DIFF
--- a/docs/recipes/running-task-steps-per-folder.md
+++ b/docs/recipes/running-task-steps-per-folder.md
@@ -40,17 +40,12 @@ gulp.task('scripts', function() {
    var folders = getFolders(scriptsPath);
 
    var tasks = folders.map(function(folder) {
-      // concat into foldername.js
-      // write to output
-      // minify
-      // rename to folder.min.js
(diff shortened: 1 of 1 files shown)


phated:
Why not above each pipe? I think that's how other recipes do it

LoganArnett:
So like this @phated 
``` javascript
gulp.task('scripts', function() {
   var folders = getFolders(scriptsPath);

   var tasks = folders.map(function(folder) {
      return gulp.src(path.join(scriptsPath, folder, '/**/*.js'))
        // concat into foldername.js
        .pipe(concat(folder + '.js'))
        // write to output
        .pipe(gulp.dest(scriptsPath)) 
        // minify
        .pipe(uglify())    
        // rename to folder.min.js
        .pipe(rename(folder + '.min.js')) 
        // write to output again
        .pipe(gulp.dest(scriptsPath));    
   });
```

phated:
Thanks!

PR #1535, opened 2016-02-25
add gulp.requireNodeFlags

rhendric:
This is a work in progress; I'd want to add documentation and tests but I wanted to toss the basic idea out there to see if there's heavy pushback first.

This provides a way for a gulpfile or a gulp plugin to specify that it requires special Node.js flags (like `--harmony_proxies`). Adding `gulp.requireNodeFlags('--harmony_proxies', ...)` to the top of a gulpfile is all that's required. It reuses the same respawn code that gets used currently when using Node.js flags as arguments to gulp, so instead of getting an error message if you forget to invoke gulp with the right things, gulp just quietly restarts with the missing flags added.

I'm torn on whether `requireNodeFlags` should go here or in `gulp-utils`; the latter feels slightly more correct if I expect plugins and not just gulpfiles to use this, but then I'd have two PRs to wrangle so I figured I'd get confirmation on that from you all first.

Feedback please?

### DIFF
--- a/bin/gulp.js
+++ b/bin/gulp.js
@@ -113,7 +113,17 @@ function handleArguments(env) {
   }
 
   // This is what actually loads up the gulpfile
-  require(env.configPath);
+  try {
+    require(env.configPath);
+  } catch (e) {
+    if (e.missingFlags) {
+      var respawn = require('flagged-respawn/lib/respawn');
+      respawn([process.argv[0]].concat(process.execArgv).concat(e.missingFlags).concat(process.argv.slice(1)));
+      return;
+    } else {
+      throw e;
+    }
+  }
   gutil.log('Using gulpfile', chalk.magenta(tildify(env.configPath)));
 
   var gulpInst = require(env.modulePath);
--- a/index.js
+++ b/index.js
@@ -40,6 +40,21 @@ Gulp.prototype.watch = function(glob, opt, fn) {
   return vfs.watch(glob, opt, fn);
 };
 
+function MissingFlagsError(missingFlags) {
+  Error.call(this);
+  this.missingFlags = missingFlags;
+}
+util.inherits(MissingFlagsError, Error);
+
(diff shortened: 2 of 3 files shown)


phated:
You can just do this in your gulpfile.  Check process.argv and throw if it doesn't exist.

rhendric:
No, that just fails the process with an error. What I'm proposing here restarts gulp with the missing flags.

phated:
3.x is API locked and we aren't inclined in expanding the 4.0 API surface, as it is nearing completion.  This would be something that would have to be handled by gulp-cli but I don't think it has a wide enough use case to warrant adding it.

rhendric:
If you're not willing to take the whole feature, is there a hook or something you'd be willing to add to gulp-cli to enable shutting gulp down gracefully? I can do just about all of this respawning stuff from a user library; the only missing piece is some way to stop gulp's main load-and-run process without literally `process.exit`ing or dumping a bunch of error noise to console.

Just some kind of try/catch around `require(env.configPath)` that returns silently as long as the error caught meets some condition would be enough for me—it doesn't have to know about missing flags, just check for an `interruptGulp` property or something. Or something else that achieves that goal; it doesn't have to be error catching. I totally understand wanting to keep a hard line against API bloat, but can you help me out at all? Without something like this, it's basically unfeasible to make gulp plugins that depend on in-progress Node.js features, because approximately zero people are going to want to call `gulp --harmony-proxies` every time they want to build.

phated:
@rhendric definitely willing to discuss things but your example of writing plugins that use this new API doesn't seem to make sense because gulp plugins aren't supposed to require gulp as a dependency.  Willing to brainstorm but I think it needs to be some subset of your ideas and I'm not sure how to handle it.  Could you open an issue on the gulp-cli repo since it will absolutely have to go there if we solve it.

rhendric:
Will do, thanks!

Issue #1485, opened 2016-01-11
[Notice] lodash deprecation warning


phated:
We are aware of the `lodash` deprecation warning upon install of `gulp` 3.x.  This is due to a dependency deep in our dependency tree that we are unable to upgrade due to API breaking changes and stability issues with newer versions.  There is nothing wrong with the dependency or that version of `lodash` for our use cases.  We will be updating to a different library altogether in gulp 4 and the message will go away.

**Please do not open issues about this topic.
Thank you,
-The gulp team**

phated:
This looks to no longer be happening.

PR #1555, opened 2016-03-08
Update graceful-fs to ^4.0.0

lipis:
I know that it's not that easy.. but at one point we'll have to update that.

```bash
npm WARN deprecated graceful-fs@3.0.8: graceful-fs version 3 and before will fail on newer node releases. Please update to graceful-fs@^4.0.0 as soon as possible.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "coveralls": "^2.7.0",
     "eslint": "^1.7.3",
     "eslint-config-gulp": "^2.0.0",
-    "graceful-fs": "^3.0.0",
+    "graceful-fs": "^4.0.0",
     "istanbul": "^0.3.0",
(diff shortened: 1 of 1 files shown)


callumacrae:
https://github.com/gulpjs/gulp/issues/1485

It would probably be useful if it were updated in the 4.0 branch, but it won't but updated in 3.x because it's a breaking change.

phated:
Yeah, we shouldn't update this dependency because it might break the dep tree (even though it only is a devDep).  Maybe we could just remove it since it is just used for tests?  @callumacrae @contra thoughts?

callumacrae:
Are there likely to be EMFILE errors in the tests if it is removed?

phated:
See the published notice.

PR #1567, opened 2016-03-17
Log require fail reason

blakeembrey:
Currently, requiring modules like `ts-node/register` can fail for unrelated issues (invalid `tsconfig.json`, no `typescript` installed, etc). It would be helpful for debugging to log out the reason it failed.

Related to https://github.com/TypeStrong/ts-node/issues/26.

It is possible to change the error level to `error.stack` instead, but the message is probably enough most of the time. Also, it may make sense to log it as part of the same log message - not sure what the etiquette is there.

### DIFF
--- a/bin/gulp.js
+++ b/bin/gulp.js
@@ -53,8 +53,9 @@ cli.on('require', function(name) {
   gutil.log('Requiring external module', chalk.magenta(name));
 });
 
-cli.on('requireFail', function(name) {
+cli.on('requireFail', function(name, error) {
   gutil.log(chalk.red('Failed to load external module'), chalk.magenta(name));
+  gutil.log(error.message);
 });
 
 cli.on('respawn', function(flags, child) {


phated:
Sorry, js variant requires are supposed to be an optional require and the message shouldn't be intrusive, like a stack would be.

blakeembrey:
@phated It's using the error message, not the stack. How does one know why it has failed otherwise? You're basically leaving it up to users to be confused and figure out why the module is failing. For example, the difference between local or globally installed.

Also, assuming you've got here it's all intentional. You've written a `gulpfile.*` in a different language, you'd damn well want to know why it's failing.

blakeembrey:
There's zero impact on someone with a normal `gulpfile.js`, and zero difference on a user that is already using custom extensions other than now you'll tell them why things are actually failing.

phated:
You already get the failure errors if you make a typo in a gulpfile.* - just not from requiring the module.  If we implement the thing you suggested, it dumps a ton of information in the babel case, among others due to fallbacks.

phated:
Oh, and the information is not useful in those cases.

phated:
Feel free to submit this to gulp-cli, where we have a real logger and you could put it as a non-default log level.

blakeembrey:
Oh, I wasn't aware `gulp-cli` was a thing. Looks like similar code. So the same PR https://github.com/gulpjs/gulp-cli/blob/master/index.js#L63?

phated:
gulp-cli is the splitting of the CLI from the gulp codebase to support multiple versions of gulp with breaking changes (3.x -> 4.0) and can ship versions outside of the gulp release cycle (following semver).  We've added a real logger with 4 log levels, so you'd have to pick one that's at a higher log-level (https://github.com/gulpjs/gulp-cli/blob/master/lib/shared/log/toConsole.js#L9-L14) so probably `debug`

Issue #1304, opened 2015-10-03
Debouncing watcher tasks with Gulp 4


ide:
With Chokidar there's no debounce option from what I've seen and instead it's up to the user of Chokidar to pass in a debounced function. So currently I do this:
```js
gulp.watch('**/*', _.debounce(compile, 200));
```

Because lodash doesn't copy the provided function's name, my Gulp log says, "[03:06:06] Starting 'debounced'..." instead of "compile". The more serious issue is that the logs never say "Finished 'debounced'..." because the `done` callback that `gulp.watch` passed to the first invocation of the debounced function is ignored.

So what I'm looking for is an easy and reliable way to debounce `gulp.watch` callbacks, possibly by adding debounce support to `gulp.watch` itself via the options arg.

mmoss:
:+1: This is something my team members have been bugging me about for a while xD

Was thinking with gulp4 I might be able to use `_.debounce(...)` to accomplish this, but unsure...

es128:
Are we referring to the `debounceDelay` option in gaze? I think that was a workaround for duplicate events (caused by editor or OS wonkiness). Chokidar tries to detect and suppress duplicate events without asking users to fiddle with settings, and considering it provides only an `EventEmitter` interface, it doesn't make sense to me to add denounce functionality there. 

The callback interface is provided directly by gulp now, so it's up to the gulp maintainers to decide whether they want to add `debounce` as a convenience option for running tasks.

es128:
What I'd expect to be much more useful than a time based debounce, however, is something that would know how to abort/restart a task upon subsequent events, or something that queues it up to run one more time after finished if more triggering events have been received while it was running.

ide:
@es128 Agreed that this option likely doesn't belong in Chokidar. It can be implemented in Gulp or a helper library.

My use case is that I have three jobs that Gulp runs: Babel, Webpack, and a Koa server. When there's a change in my source files, `gulp.watch` tells Babel to run (Webpack also independently has its own watcher), which produces new files in my build directory. Another `gulp.watch` command notices the newly built files and restarts the Koa server. The issue is that Babel and/or Webpack sometimes output multiple files within a short time span and the Koa server restarts several times.

So I see two solutions for my project:
- Debounce the `gulp.watch` handler that restarts Koa. This needs to debounce across multiple files (e.g. if Webpack writes a.js, b.js, and c.js, Koa should restart just once or twice depending on the policy for debouncing.)
[...]

PR #1569, opened 2016-03-17
add debounce feature

hiroqn:
#1304 

need `opt.wait` option for debounce in watch function
because, `fn` is wrapped `gulp.parallell`
so we cant use like this.

```js
gulp.watch(glob, debounce(gulp.series(scripts, unitTests), 200));
```

(https://github.com/gulpjs/gulp/blob/4.0/index.js#L42)

```js
  if (typeof task === 'function') {
    fn = this.parallel(task);
  }
```
 

or no wrap `fn` `gulp.parallell`

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ var util = require('util');
 var Undertaker = require('undertaker');
 var vfs = require('vinyl-fs');
 var chokidar = require('chokidar');
+var debounce = require('debounce');
 
 function Gulp() {
   Undertaker.call(this);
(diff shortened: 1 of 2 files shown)


TrySound:
@hiroqn `gulp.watch` is `chokidar`. Try `atomic` option https://github.com/paulmillr/chokidar#errors . Usually debouncing is not required.

phated:
We aren't adding debounce directly to gulp.  Please see https://github.com/gulpjs/gulp/issues/1304#issuecomment-177631437 for more information and a POC that I built.

hiroqn:
@TrySound 
I read and chokidar and in my opinion, atomic option is for one same file, 

@phated 
I already saw it ...
but queueing system do not solve some problem like we encount

phated:
@hiroqn then help me work on the proof of concept instead of sending a PR like this that we already stated would be rejected

PR #1574, opened 2016-03-21
Use gulp-cli in Getting Started

hudochenkov:
I've noticed this part is outdated, so I took this change from `master`.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -2,8 +2,11 @@
 
 #### 1. Install gulp globally:
 
+__If you have previously installed a version of gulp globally, please run `npm rm --global gulp`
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -1,9 +1,10 @@
 # Getting Started
 
-#### 1. Install gulp globally:
```
callumacrae: Why have you deleted the heading?

hudochenkov: @callumacrae i'm so sorry, it was deleted by mistake. Thank you for noticing.

phated:
Thanks.  We definitely need to do another pass and update the 4.0 docs.

PR #1591, opened 2016-04-05
Add undocumented event.type

kdex:
This event type is fired when renaming a file.


phated:
Thanks.  I'm pretty sure this event is platform dependent and won't always be emitted.  Caution advised.

PR #649, opened 2014-08-31
add `run-grunt-tasks-from-gulp` recipe

geekflyer:
link to #646

Issue #646, opened 2014-08-30
Add recipe to integrate grunt tasks into gulp


geekflyer:
Hi, I just found with the latest grunt version it's possible to run tasks without the grunt cli and without a gruntfile. That makes it possible to run grun tasks within gulp relatively seamlessly (and without a plugin) and I was able to remove `gulp-grunt` from our project, which caused some issues.

**I'm just wondering if you would accept a pull request for a recipe on that topic (before I invest my time).**

Here's some more info on this new grunt option:
https://github.com/gruntjs/grunt/commit/77155e3f61d213c529a16ea595fc486fdeca28c9

phated:
@geekflyer this looks like it shipped with grunt 1.0 - did you want to resubmit?  If not, I will close this soon.

phated:
Completed by #1592

PR #1592, opened 2016-04-05
add `run-grunt-tasks-from-gulp` recipe

geekflyer:
resubmit of #649 now that grunt 1.0.0 is released on npm as suggested in #646. @phated


phated:
Any idea why I'd be getting the following error message when running the example?
```
Running "jasmine:all" (jasmine) task
Warning: glob pattern string required Use --force to continue.
```

geekflyer:
Hey, thanks for noticing. After upgrading grunt jasmine I'm also getting this error - even when attempting to run the grunt task directly via Grunt CLI and Gruntfile (no gulp involved). Tried to hack around it a bit, but it seems some weird issue / incompatibility with grunt-contrib-jasmine@1.0.0 and grunt@1.0.0 in combination. I couldn't even get the most simple example from the grunt-contrib-jasmine running. Anyways - since this recipe is not about grunt-contrib-jasmine I just simplified the example to use grunt-contrib-copy which works fine. Not the most meaningful example but I think it gets to the point.

phated:
Excellent!  Thanks @geekflyer

PR #1607, opened 2016-04-18
Update getting-started.md

zyf0330:


phated:
The 4.0 docs are a work-in-progress and none of them should be taken verbatim.  While this might help in the short-term, it would be tedious to go through and remove them if notes like this were added everywhere.

Issue #1571, opened 2016-03-17
[Notice] graceful-fs deprecation warning


phated:
We are aware of the graceful-fs deprecation warning upon install of gulp 3.x.

This is due to:
1. A transitive graceful-fs devDependency

We are unable to upgrade due to API breaking changes.

There is nothing wrong with the dependency, especially since it is only used in development. We will be updating or removing it in gulp 4 and the message will go away.

**Please do not open issues about this topic.
Thank you,
-The gulp team**

phated:
This should be no longer an issue due to the amazing work by @thealphanerd at https://github.com/isaacs/node-graceful-fs/commit/07701dc98bc088e1a71f18a8f59a970e3354646f

I'll leave this open until I have a chance to verify.  Unlocking incase anyone has a chance to test it out.

piotr-cz:
how to test?

phated:
Looks like there was a follow up at https://github.com/isaacs/node-graceful-fs/commit/07701dc98bc088e1a71f18a8f59a970e3354646f#commitcomment-18659275

Locking again with the same conclusion.  Sorry everyone.

phated:
As per https://github.com/gulpjs/gulp/pull/1760#issuecomment-242213701 - It looks like I can close this now that graceful-fs 3.0.10 has been released.

phated:
Looks like I need to reopen this 😦

phated:
gulp 4 has been published under the `next` tag (`npm install gulp@next`) so I'm going to close this.

PR #1608, opened 2016-04-18
3.9.2

bradchesney79:
So, the vinyl-fs dependency is behind the times quite a bit. I haven't had time to test this, but this would potentially remove the warnings of many npm gulp users...


callumacrae:
https://github.com/gulpjs/gulp/issues/1571

PR #1609, opened 2016-04-18
4.0 Readme: Update incremental builds example

chmln:
`paths.images` in example isn't seen anywhere.
I've added a simple relevant example to avoid any confusion.


phated:
thanks!

PR #1615, opened 2016-04-23
Bottom example is now runnable

thatchej:
Previously, the app would throw the error:

"Task 'default' is not in your gulpfile"


thatchej:
Could also change Line 32 to read:

`gulp.watch(['src/*.js', 'test/*.js'], ['mocha']);`

phated:
You changed the glob.  Might as well leave it the same.

phated:
Thanks!

PR #1632, opened 2016-05-03
Add note for compatibility with gulp 3.x 

beenanner:
Reference to comment in #1445 - adding note to let folks know this only works with the latest published gulp-cli version


phated:
Thanks!

PR #1653, opened 2016-05-12
correct gulp.task signature

makoConstruct:
`gulp.task(name [, deps, fn])` means that only `gulp.task(name, deps, fn)` and `gulp.task(name)` are allowed, which is apparently not the case and it's a bit confusing. The correct notation, `gulp.task(name [, deps] [, fn])`, also permits for `gulp.task(name, deps)`, and `gulp.task(name, fn)`.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -3,7 +3,7 @@
 Jump to:
   [gulp.src](#gulpsrcglobs-options) |
   [gulp.dest](#gulpdestpath-options) |
-  [gulp.task](#gulptaskname--deps-fn) |
+  [gulp.task](#gulptaskname--deps--fn) |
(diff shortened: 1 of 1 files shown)


phated:
@makoConstruct thanks for the update; however, you must change the Table of Contents link at the top (I believe) because the updated header changes the way GitHub sets the ID on it.

phated:
Thanks!

skyl:
What about the little description option? `gulp.task(name, descr, fn)` is also possible? I'm actually looking to have a description and dependencies but I can't find the signature for that offhand. Doesn't seem to support `gulp.task(name, deps, descr, fn)`

phated:
@skyl your comment doesn't belong here but I think you are looking for https://medium.com/gulpjs/gulp-sips-custom-task-metadata-9a2dc80ac7b1

PR #1661, opened 2016-05-14
Fix Gulp4 docs browsersync example

limdauto:
@phated I'm terribly sorry there was a small mistake in the version I submitted yesterday. Hope you don't mind merging another commit.


phated:
Good catch.  Totally missed those.

PR #1663, opened 2016-05-14
Update CI

adam-moss:
Update Travis with the latest version of node, major version releases of node, and iojs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 sudo: false
 language: node_js
 node_js:
-  - "0.10"
-  - "0.12"
+  - "6"
+  - "5"
   - "4"
-  - "stable"
+  - "0.12"
+  - "0.10"
 after_script:
   - npm run coveralls
 git:


phated:
Our only supported platforms will be node 0.10, 0.12, 4, 5 and 6; we will not be supporting specific version of node like 4.2 nor iojs releases.

adam-moss:
Updated as per comment.

.travis.yml
```diff
@@ -1,10 +1,12 @@
 sudo: false
 language: node_js
 node_js:
-  - "0.10"
-  - "0.12"
+  - "node"
```
phated: no need for `node` if all the versions are specified.  We want to be explicit when we support something new (e.g. node 7 when it comes out) but right now this just does another run of node 6.

adam-moss:
Updated :)

phated:
@adam-moss please remove the `node` line.  Since we are being specific about the versions we are supporting on CI, I don't want a replacement for `stable`.  Thanks

adam-moss:
Done.

phated:
Closed by #1692

PR #1688, opened 2016-06-09
graceful-fs v3.0.0 related deprecation warning (node v6.x.x)

unidwell:
Hi,

when using gulp 3.9.1 with shell autocompletion, node v6.x.x reports deprecation message related to graceful-fs v3.0.0 (on each autocompletion):

`gulp (node:36477) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.`

In order to do the update in the current gulp version, packages using old graceful-fs need to be replaced or updated. Upgrading to the latest vinyl-fs version implies 'back-porting' vfs.watch functionality.

This simple change will give gulp v3.x.x a little bit longer lease on life until v4.0.0. But the possible breaking change could be in the fact that change involves glob-watcher, which uses Chokidar in it's latest version and not all watch options are probably backward compatible. Not sure, how much that is an issue in practice. 

All best,
Uni

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ var Orchestrator = require('orchestrator');
 var gutil = require('gulp-util');
 var deprecated = require('deprecated');
 var vfs = require('vinyl-fs');
+var watch = require('glob-watcher');
 
 function Gulp() {
   Orchestrator.call(this);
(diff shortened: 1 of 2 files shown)


phated:
There's an issue detailing why we don't do this.

unidwell:
Tnx, found the mentioned issue.

PR #1690, opened 2016-06-14
Add "Rollup with rollup-stream" recipe.

Permutatrix:
The code is essentially the same as the first two examples from rollup-stream's README.md. Once [rollup-plugin-gulp](https://github.com/Permutatrix/rollup-plugin-gulp) earns the right to exist by having unit tests, I'll add a recipe for its use together with rollup-stream. That's assuming rollup-plugin-gulp turns out to be a good idea and not a horrifying aberration, of course.


docs/recipes/rollup-with-rollup-stream.md
```diff
@@ -0,0 +1,62 @@
...
+    // buffer the output. most gulp plugins, including gulp-sourcemaps, don't support streams.
+    .pipe(buffer())
+
+    // tell gulp-sourcemaps to load the inline sourcemap produced by rollup-stream.
+    .pipe(sourcemaps.init({loadMaps: true}))
+
+        // transform the code further here.
+
+    // if you want to output with a different name from the input file, use gulp-rename here.
+//  .pipe(rename('index.js'))
```
phated: indentation is off here.  Should this just be removed?

Permutatrix: I put the `//` at the start of the line so it would line up with the other `.pipe`s. I'll change it if you want, but I prefer it this way myself.

phated: It looks really weird when I view the rich diff.  Align it with the `// transform the code further here.` line

Permutatrix:
Alright, the indentation has been adjusted.

PR #1692, opened 2016-06-15
testing under node version 5 and 6

MikailBag:


ericmdantas:
I'd keep stable, though.

phated:
Stable is deprecated. We are removing it everywhere. Gulp 3.x won't be supported on node 7

PR #1739, opened 2016-07-31
Fix Error logging

santi6291:
The on error docent catch the Browserify compile errors causing Watch task to break


phated:
We don't want to show everything in each recipe.  There's already a note at the beginning that links to the recipe about handling errors.

Issue #1640, opened 2016-05-05
Old graceful-fs version in gulp 3.x


ChALkeR:
Opening this as a separate issue, as #1571 got comments closed (and describes a slightly different issue), and my comment in #1604 got moderated without a notice.

`graceful-fs` below version 4.0 is going to break on Node.js 7.x (which is going to happen in about 5 months from now).

_If you are sure that everything is covered and understood correctly — just close this, please :-) (and excuse me for opening a new issue), I am just making sure that nothing is missed here._

Your assessment of this issue in #1571 is not precise:
> There is nothing wrong with the dependency, especially since it is only used in development. We will be updating or removing it in gulp 4 and the message will go away.

There are two warnings there — a warning on installation of graceful-fs lower than 3.x and a warning in runtime when an old graceful-fs version does unsupported stuff (which it does upon `require`).

I am only talking about the runtime warning here that comes from Node.js. That runtime warning is emitted only in those situations which would be turned into throws on Node.js 7.0 and above.

Now, just running `gulp` even  with an empty `gulpfile.js` file emits this runtime warning:
```
(node:25661) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
```
which is going to turn into a throw in Node.js 7.0.

The dependency tree to `graceful-fs` looks like this:
```
└─┬ gulp@3.9.1 
  └─┬ vinyl-fs@0.3.14 
    ├─┬ glob-watcher@0.0.6 
    │ └─┬ gaze@0.5.2 
    │   └─┬ globule@0.1.0 
    │     └─┬ glob@3.1.21 
    │       └── graceful-fs@1.2.3 
    └── graceful-fs@3.0.8 
```

`graceful-fs@1.2.3` is not reevalating native module sources, so it doesn't trigger that warning and would not be affected by that throw in Node.js 7.0.

Only `graceful-fs@3.0.8` causes that warning here, which is required only by `vinyl-fs@0.3` (apart from being a devDependency). `vinyl-fs@0.3` is loaded both by `orchestrator` (which is loaded only by `gulp`) and directly by `gulp`.

To my knowledge, there shouldn't be breaking changes in `graceful-fs` update between 3.0 and 4.0 and that should be a clean update.

A straightforward fix for to this would be to release a [vinyl-fs](https://github.com/gulpjs/vinyl-fs) 0.3.x with `graceful-fs@4`, perhaps.

ChALkeR:
[...]

PR #1760, opened 2016-08-18
POC: Introduce Node.js 6+ compatibility mode for gulp@3

ChALkeR:
Refs: https://github.com/gulpjs/gulp/issues/1640#issuecomment-240683358

This is unfinished POC, it lacks the documentation and perhaps the tests.

It works though, and successfully unbreaks gulp@3 with Node.js 6+ if gulp was required using `require('gulp/compat')`.

This is a pure opt-in, it does not change anything unless the user specifically enables it (except for installing one extra unused package).

If gulp is already loaded at the time of the opt-in, then it throws an error —  so that deps won't misuse it and change the behaviour of everything, — this makes this option available only to the top-level gulpfile itself.

If this would be accepted, I will gladly move [vinyl-fs-03-compat](https://github.com/ChALkeR/vinyl-fs-03-compat) to gulpjs both on GitHub and npm.

/cc @phated, @contra, @TheAlphaNerd, @jasnell

Thoughts?

### DIFF
--- a/compat.js
+++ b/compat.js
@@ -1,0 +1,8 @@
+'use strict';
+
+// Initialize vinyl-fs-wrap in Node.js 6+ compatibility mode
+var vfsWrap = require('./vinyl-fs-wrap');
+vfsWrap.initCompat();
+
+// Export the origial Gulp
+module.exports = require('./index');
--- a/index.js
+++ b/index.js
@@ -4,7 +4,10 @@ var util = require('util');
 var Orchestrator = require('orchestrator');
 var gutil = require('gulp-util');
 var deprecated = require('deprecated');
-var vfs = require('vinyl-fs');
+
+var vfsWrap = require('./vinyl-fs-wrap');
+vfsWrap.init();
+var vfs = vfsWrap.vfs;
 
 function Gulp() {
   Orchestrator.call(this);
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "semver": "^4.1.0",
     "tildify": "^1.0.0",
     "v8flags": "^2.0.2",
-    "vinyl-fs": "^0.3.0"
+    "vinyl-fs": "^0.3.0",
+    "vinyl-fs-03-compat": "^0.3.15"
   },
   "devDependencies": {
     "coveralls": "^2.7.0",
--- a/vinyl-fs-wrap.js
+++ b/vinyl-fs-wrap.js
@@ -1,0 +1,30 @@
+'use strict';
+
+var compat = false;
+
+function init() {
+  if (exports.vfs) {
+    return;
+  }
+  exports.vfs = require('vinyl-fs');
+}
+
+function initCompat() {
+  if (exports.vfs) {
+    if (!compat) {
+      throw new Error(
+        'Gulp was already initialized without Node.js 6+ compatibility mode!\n' +
+        'Make sure that you require gulp/compat before other gulp plugins.\n' +
+        '\n' +
+        'Note that dependencies should not force this mode.\n' +
+        'The compatibility mode is intended only for the top-most gulpfile.js.\n'
+      );
+    }
+    return;
+  }
+  compat = true;
+  exports.vfs = require('vinyl-fs-03-compat');
+}
+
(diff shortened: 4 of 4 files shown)


TheAlphaNerd:
Would it potentially make sense to wrap the require of `graceful-fs` in a try-catch. If the require fails then opt to using the compat version.

This could be a good way to allow for a transparent fix for those on affected versions.

phated:
I find this an interesting workaround. Let me think on it for awhile.

ChALkeR:
@phated Would it be ok to transfer vinyl-fs-03-compat now? Asking in advance because I don't want to spam you =).

phated:
@ChALkeR no worries.  If we decide to pull this workaround in, I'll create a branch/tag in vinyl-fs proper

ChALkeR:
@phated Ah, ok. That makes sense =).

jasnell:
just fyi, related to this: https://github.com/nodejs/node/pull/8166. It doesn't solve the issue but it's one small step.

ChALkeR:
Another note: this is not a separate module, but a mere separate entry point for these reasons:
 1. That requires less changes for the users to integrate, it's a one-liner this way.
 2. If any dependencies used in the gulpfile also do `require('gulp')` on their own for some reason — this top-level opt-in would propogate to those.
 3. Introducing a separate module would likely raise more concerns and recelive less usage.
 4. It would be harder to support if this was a separate package.

ChALkeR:
Ah, this approach has a drawback.

The propogation would only work if the `gulp` loaded by `require('gulp')` is the exact same package, not a separate copy in a deeper `node_modules` dir. This could be fixed by making the `compat` flag in `vinyl-fs-wrap.js` stored per-process (i.e. in some global storage), not per-gulp-module.

Thoughts?

mgol:
> The propogation would only work if the `gulp` loaded by `require('gulp')` is the exact same package, not a separate copy in a deeper `node_modules` dir.

Is this even supported by task managers like Gulp?

> This could be fixed by making the `compat` flag in `vinyl-fs-wrap.js` stored per-process (e.g. in `global`), not per-gulp-module.

Had this been necessary, could we use a symbol to minimize the possibility of a collision?

ChALkeR:
@mgol 
> Is this even supported by task managers like Gulp?

I could definitely imagine people doing `require('gulp')` in some package out there for some reason. And I could imagine such package being required from the gulpfile, for some reason. It doesn't even matter if that code path that actually uses the `require('gulp')` result would be not taken in the dependency — `require('gulp')` would crash itself.

Note that about 2943 packages depend on `gulp` via the `dependencies` chain (even _without_ the `devDependencies`). That's about 1% of all npm packages.

ChALkeR:
@mgol e.g. [sifttt](https://www.npmjs.com/package/sifttt) is both managed using a gulpfile with a top-level `require('gulp')` and does `require('gulp')` inside the package itself: [markbirbeck/sifttt/blob/master/lib/beam/GulpSource.js#L3](https://github.com/markbirbeck/sifttt/blob/master/lib/beam/GulpSource.js#L3)

ChALkeR:
Note: graceful-fs@3.0.10 was released by @isaacs and that version does not trigger the warning anymore, so this can be closed now.

ChALkeR:
Closed, will reopen if any issues arise with the new graceful-fs or if the change there gets reverted for some reason.

phated:
Works for me.  Thanks to everyone that contributed towards a solution!

ChALkeR:
For posterity and anyone else with gracefsul-fs@3: this was done with https://github.com/isaacs/natives/blob/master/index.js, which also uses a bunch of undocumented and internal things, relies on implementation details, and also has some other hacks (e.g. it's basically a noop on the `buffer` module, and that one was just the module that will _definitely_ have issues with that approach, and other modules also could be affected, e.g. I suspect `util` is affected).

That _perhaps_ shouldn't manifest itself in the near future in regards to graceful-fs@3, (hence it was ok to close this now and leave things as they are), but in the long term noone should ever user graceful-fs below v4 and noone should ever use that `natives` module in their own projects.

PR #1762, opened 2016-08-18
POC: Fallback to graceful-fs@4 if graceful-fs@3 is unsupported

TheAlphaNerd:
This uses a vinyl-fs@0.3 version coupled with graceful-fs@4 as a
fallback for the regular combination of vinyl-fs@0.3 and graceful-fs@3,
so that gulp@3 doesn't break on newer Node.js versions where
graceful-fs@3 is unsupported.

Current setups and Node.js versions (6.x and below) are not affected.

Alternative approach to #1760

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,13 @@ var util = require('util');
 var Orchestrator = require('orchestrator');
 var gutil = require('gulp-util');
 var deprecated = require('deprecated');
-var vfs = require('vinyl-fs');
+
+var vfs;
+try {
+  vfs = require('vinyl-fs');
+} catch (e) {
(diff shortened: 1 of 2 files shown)


phated:
I think this is the wrong approach. I know you want to attempt to handle this transparently but it would change the semantics of the runtime without any way to know. Let's continue discussion but I'm going to close so it doesn't accidentally get merged.

TheAlphaNerd:
@phated would you be open to a cli flag? 

`gulp --node-compat`?

phated:
Not really. We don't support forcing certain flags (yet) which makes that clumsy to use. So far I like the separate module idea but I need to think on it.

PR #1775, opened 2016-08-29
Illustrate use of gulp-load-plugins | Illustrate use of caching

Nitin-Prabhakar:
Add a style task , implementing caching in gulp.
Illustrate gulp-load-plugins instead of manually requiring all available plugins


Nitin-Prabhakar:
Needs some more modifications. closing this to make them and re submit

PR #1777, opened 2016-08-29
Illustrate use of gulp-load-plugins | Illustrate use of caching

Nitin-Prabhakar:
Added a caching behaviour with example
Added gulp-load-plugins to autoload plugins and avoid manual requires
Defined the anatomy of a gulp task - by documentation
Described how to create a function that is used by a task

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,17 +28,44 @@ This file will give you a taste of what gulp does.
 
 ```js
 var gulp = require('gulp');
-var coffee = require('gulp-coffee');
-var concat = require('gulp-concat');
-var uglify = require('gulp-uglify');
-var imagemin = require('gulp-imagemin');
-var sourcemaps = require('gulp-sourcemaps');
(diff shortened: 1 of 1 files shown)


alexeyraspopov:
> Added gulp-load-plugins to autoload plugins and avoid manual requires

Maybe `gulp-load-plugins` can be just mentioned below? It's really opinionated approach to import plugins.

Nitin-Prabhakar:
>Maybe gulp-load-plugins can be just mentioned below? It's really opinionated approach to import plugins.

Hmmm. Well we can. but I thought, ease of development takes priority and gulp-load-plugins saves a lot of space and forgives if we forget to require something. Saves debug time in such cases.

Do the other changes look okay though?

phated:
None of this belongs in the main readme. We have things like recipes and docs that can cover these topics.

PR #1778, opened 2016-08-29
Create Responsify-Images.md

retrocausal:
Add a recipe to illustrate generation and minification of images for the web


Nitin-Prabhakar:
@QWp6t 

Agreed. Syntactically, the code in a recipe must align with its predecessors. made subtle changes here and there, but took care of dropping ```gulp-load-plugins```, and replaced ```$``` to ```gulp```
Also replaced ```$$```(previously aliasing gulp-load-plugins) to actual variable names as in their require declaration

phated:
We are currently rewriting all the docs so I'm going to close this.

PR #1813, opened 2016-09-20
Update browserify-uglify-sourcemap.md

m98:
It is better for beginners to clarify the file (code) is inside gulpfile.js


phated:
This looks fine.

Issue #4, opened 2013-07-06
check for gulpfile case insensitive


phated:
think grunt added this in 0.4.1, it is a nice to have.

PR #1827, opened 2016-10-06
Update to v4 of graceful-fs

magnudae:
With v4, graceful-fs is no longer monkey patched. 
The old version of graceful-fs will also be deprecated when node >= 7 
Still there are several, much downloaded, packages around which has to update aswell. 

Lint and tests are good. 

[Ref v4](https://github.com/isaacs/node-graceful-fs#v4)

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "coveralls": "^2.7.0",
     "eslint": "^1.7.3",
     "eslint-config-gulp": "^2.0.0",
-    "graceful-fs": "^3.0.0",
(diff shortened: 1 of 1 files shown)


contra:
We've already done this on the latest vinyl-fs. You can update to gulp 4.0 (`npm install gulpjs/gulp#4.0`) or use the latest vinyl-fs directly.

phated:
See also https://github.com/gulpjs/gulp/issues/1571

magnudae:
Great. 
Looks like I overlooked the issue there. My bad.

PR #1841, opened 2016-10-21
Fix old code

tyscorp:
According to npm statistics, 0.01% of node.js users are still using 0.10. I think it's now safe to update the codebase a bit to reflect the community.

Thanks!

### DIFF
--- a/bin/gulp.js
+++ b/bin/gulp.js
@@ -27,7 +27,7 @@ var cli = new Liftoff({
 
 // Exit with 0 or 1
 var failed = false;
-process.once('exit', function(code) {
+process.once('exit', code => {
   if (code === 0 && failed) {
     process.exit(1);
   }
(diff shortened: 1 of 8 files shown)


terinjokes:
At this time, since several distributions are still shipping Node.js 0.10, we're continuing to support this version.

phated:
Those are straight incorrect numbers.  Also, if you think that littering your code with arrow functions is helping, you should think again.

phated:
Non-bullshit numbers from https://twitter.com/seldo/status/781258069678391296

> Node 0.10 still has 13% share; don't drop support yet.

node10_sep_16

callumacrae:
"Fix" old code? Was it broken?

Fat arrow functions are less likely to be understood by junior developers or people who don't know ES6. This isn't just about Node 0.10, this is about understandable and accessible code, too. I notice you've left the documentation alone, though.

PR #1850, opened 2016-11-05
bump dependencys

frank-dspeed:


phated:
These are breaking changes.

frank-dspeed:
then we need to fix them till it works!

PR #1866, opened 2016-12-01
adding another article

mmaksimovic:


phated:
This article has a few things wrong that we've stopped promoting, including: globally installing gulp instead of gulp-cli, not returning streams, gulp-util usage, and using blacklisted plugins.

Going to close this.

PR #1891, opened 2017-01-25
upgrade vinyl-fs

sh977218:
fix npm warning:
graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "semver": "^4.1.0",
     "tildify": "^1.0.0",
     "v8flags": "^2.0.2",
-    "vinyl-fs": "^0.3.0"
+    "vinyl-fs": "2.4.4"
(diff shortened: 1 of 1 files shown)


contra:
The update to vinyl-fs will land in gulp 4.0 since it is a breaking release, the API changed significantly so it doesn't belong on the 3.x branch.

popomore:
@contra we have security issue about this version of `vinyl-fs` https://snyk.io/test/github/eggjs/egg-sequelize/7f18b0fb7d6ec4049726064d8074e1ac67e2c62b?tab=dependencies

PR #1894, opened 2017-01-31
Replace tags with keywords in package.json

Mordax:
Hey there, I noticed that gulp's info on npmjs has no keywords. 

npmjs

Thought I'd do an easy fix for you by just switching out tags for keywords.

Hope you have a great day!


contra:
Weird, don't know how that happened. Thanks!

PR #1900, opened 2017-02-02
Added note to install command

Tal-Daniel:
After I ran into trouble installing gulp, I added an example how to run as admin, on linux. 
Fixed after ran into trouble on MAC OS.


phated:
That is poor advice and you should fix your npm global permissions.

PR #1901, opened 2017-02-02
Let's update the installation command

revelt:
Let's update the installation with the correct command for v.4, applicable right now. When v.4 will be officially released, we can amend it accordingly. But for now, this is the correct installation command. Hopefully this will save people's time googling how to install.


phated:
I don't really want to do this. All the documentation for 4 is incomplete.

revelt:
You're right, we should get the whole thing done properly. I'll try to help all I can.

PR #1909, opened 2017-02-08
might be good to note to remove both packages

TerribleDev:
its not inherently wrong to just ask people to remove gulp, but it is also not a bad thing to ask them to remove both packages. Especially since they are both installed in this tutorial, and could remove any lingering issues people have


phated:
I don't see a reason to remove gulp-cli

PR #1922, opened 2017-03-10
gulp.src example '!' has wrong result

steelywing:
The example should only match `a.js`, it look like `glob-stream` will ignore the `!` at last. I tested it on Windows.


steelywing:
It should be a bug, is not doc wrong.

PR #1934, opened 2017-04-02
Use gulp-babel & babel-preset-env to set up Gulp with Babel

zhe:
Hi, I think we might need to update README about how to use Gulp with Babel. From the latest [Babel docs](https://babeljs.io/docs/setup/), we can just need to install `gulp-babel` and `babel-preset-env`. So I made a PR to represent this update. Hope someone could take some time to help and look into this. Thanks.


phated:
You are mistaken the point of those docs.  It's not to use gulp-babel in a pipeline, it's to transpile gulpfiles written in babel.

PR #1948, opened 2017-04-28
Fix issue with formatting

simonmeusel:


docs/writing-a-plugin/dealing-with-streams.md
```diff
@@ -8,7 +8,7 @@
 
 ## Dealing with streams
 
-Let's implement a plugin prepending some text to files. This plugin supports all possible forms of file.contents.
```
phated: This is correct, though file.contents should probably be wrapped in ` like `file.contents`

simonmeusel:
Ok, i updated that.

phated:
Thanks!

PR #1953, opened 2017-05-11
bump version of vinyl-fs

racheldonovan:
current version specified has a security vulnerability in a dependency, upgrading version should resolve the issue.


phated:
Please use the search. This is a breaking change.

PR #1954, opened 2017-05-11
Fixes update event

fleeco:
In order for watchify to work correctly it requires for the function to be implicitly returned.

This took a while for me to figure out, but can confirm that the change works on watchify 3.9.0

### DIFF
--- a/docs/recipes/fast-browserify-builds-with-watchify.md
+++ b/docs/recipes/fast-browserify-builds-with-watchify.md
@@ -30,7 +30,9 @@ var b = watchify(browserify(opts));
 // i.e. b.transform(coffeeify);
 
 gulp.task('js', bundle); // so you can run `gulp js` to build the file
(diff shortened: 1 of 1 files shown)


phated:
What you've changed is the same thing except you added an anonymous function.  What are you trying to achieve?

fleeco:
Going to be honest, no idea why it doesn't work without the anon function.  Ran into somebody else who had the same problem on SO and low and behold the anon function fixes the issue.

Tested it a few times and can confirm the fix works.

phated:
I'd like this to be better investigated before merging.  The change shouldn't be changing anything that would cause 1 to work and the other to not.

fleeco:
My machine was possessed by ghosts and I'm an idiot.

You can close this - after a reboot everything seems to be working.

phated:
Alright. Glad the reboot worked.

PR #1955, opened 2017-05-15
Fix rule numbers

tamj0rd2:
At the moment, rule numbers are showing as 1, 1, 1... etc

### DIFF
--- a/docs/writing-a-plugin/guidelines.md
+++ b/docs/writing-a-plugin/guidelines.md
@@ -5,45 +5,45 @@
 [Writing a Plugin](README.md) > Guidelines
 
 1. Your plugin should not do something that can be done easily with an existing node module
-  - For example: deleting a folder does not need to be a gulp plugin. Use a module like [del](https://github.com/sindresorhus/del) within a task instead.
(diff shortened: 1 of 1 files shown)


contra:
@tamj0rd2 Are these not rendering correctly? According to the github markdown spec the # doesn't matter, so using 1 repeatedly is best to ensure they always increment and don't become unordered

phated:
That's how it's supposed to work but it looks like they change or broke their parser (I'm seeing this problem everywhere someone mentions using 1 repeated).  @contra do you think we should merge this or wait to see if github fixes the bug (assuming it's not an intentional change).

phated:
It seems that github did change their parser a while back (as per https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown and https://githubengineering.com/a-formal-spec-for-github-markdown/).  @tamj0rd2 Can you open an issue on https://github.com/github/cmark to see if this is a breaking change they intended before we merge this?

tamj0rd2:
@doowb Thanks for spotting that. I've updated my commit.

phated:
Thanks!!

PR #1968, opened 2017-06-08
Modify using-multiple-sources-in-one-task.md

warvenSniper:
merge-stream do not work and cause the error: `TypeError: Cannot read property 'once' of undefined`. But event-merge work right


phated:
No, that module is unmaintained

PR #1974, opened 2017-06-17
Autobind the `task` function

tbranyen:
Bit nervous opening this, but I think it'd really improve my code, so worth a shot.

Autobinds `task` to allow for configs to be written like:

``` js
const { task } = require('gulp');

task('someTask', () => {
  console.log('here');
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,10 +8,11 @@ var vfs = require('vinyl-fs');
 
 function Gulp() {
   Orchestrator.call(this);
+  // Allow `task` to run independently of `gulp`
+  this.task = Gulp.prototype.add.bind(this);
 }
 util.inherits(Gulp, Orchestrator);
 
-Gulp.prototype.task = Gulp.prototype.add;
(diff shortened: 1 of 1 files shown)


phated:
I think this is already done in 4.0 

The 3.x branch is pretty frozen as we push towards release.

tbranyen:
Sweet! Looking forward to it, I'll just publish a scoped package until then, thanks!

phated:
Have you tried exports as task registration with newer versions of the CLI? It really cleans things up

tbranyen:
This particular tool uses Gulp under the hood to orchestrate tasks over a lerna repo, API-wise locked to the ~3 version unless 4 is backwards-compat for task flow events.

phated:
@tbranyen we emit events but they are more "standard" than orchestrator.  I believe they are `start`, `stop` and `error`

phated:
@tbranyen Sorry we can't land this in 3.x. I'm going to close.

PR #1984, opened 2017-07-10
Fix a broken markdown

HyunSeob:


phated:
Thanks!  There's tons of these lying around since github tightened their parser.

Issue #1719, opened 2016-07-08
DoS vulnerability 


aaronroberson:
Node Security (nodesecurity.io) is reporting a vulnerability with the minimatch package this module implicitly depends on:

https://nodesecurity.io/advisories/118

callumacrae:
https://github.com/gulpjs/gulp/issues/1571

Don't install gulp on your production servers.

contra:
@callumacrae No, not true - there's absolutely no case where this would ever be a problem with gulp.

I'm all for fixing security issues (literally used to be my job!) but this nodesecurity stuff is ridiculous.

You could say the same "DoS exploit" about `array.forEach`:

WARNING! `array.forEach` DoS exploit! Beware! Proof of concept:

```js
var genstr = function (len) {
  var result = [];
  for (i=0; i<=len; i++) {
    result.push(0);
  }
  return result;
}

var exploit = genarr(10000000000000000000);

// wow, incredible exploit.
console.log(“starting forEach”);
exploit.forEach(function() {
  // haha owned! omg!
})
console.log(“finishing epic exploit”);
```

Jokes aside, don't worry about this. If you have some crazy case where you're taking unsanitized user input and passing it directly into gulp.src you have other serious issues on your hands.

erikvold:
Why not just update the dependency version?

phated:
@erikvold it's called Semver.

contra:
@erikvold There are breaking changes that would surface in our API, which means we would break our API and break other people's perfectly fine code to solve a non-problem. Our next breaking release uses the latest version, you can use that if the warning really bothers you.

PR #1994, opened 2017-07-18
update vinyl-fs dependency

erikvold:
Currently the package.json dependency version is `"^0.3.0"` which results in downloading `vinyl-fs@0.3.14` which is many versions and years old now.


erikvold:
ah @plated made me think this was a semver issue https://github.com/gulpjs/gulp/issues/1719#issuecomment-315985627 but it is not, this was a waste of time sorry.

PR #1995, opened 2017-07-18
Build: CI test under node version 7 and 8

erikvold:


phated:
Looks good.  We're actually going to be removing the odd versions soon (they are always considered unstable).

phated:
Thanks!

PR #2016, opened 2017-09-03
Paraquire - paranoidal require

nickkolok:
I'm at hard work on [`paraquire`](https://github.com/nickkolok/paraquire) - security-oriented `require` alternative and I invite you to join.
This PR adds security jails to most libraries used in `gulp`; unfortunately, I haven't managed to give correct permissions to `vinyl-fs`, so it remains usually `require`d.

Should I prepare same PR to 4.0 branch?

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ var Orchestrator = paraquire('orchestrator', {
   builtin: ['events', 'util'], process: ['hrtime'],
 });
 var gutil = require('gulp-util');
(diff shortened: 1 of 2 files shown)


contra:
Cool project, not going to use it in gulp though. No tangible gains as far as I'm concerned, but I could see this being useful in other projects.

nickkolok:
Thanks for quick reply!
Could you advice another projects which I should go to?

PR #2020, opened 2017-09-24
Use meaningful processTitle for better monitoring

nguymin4:
At the moment when we run several gulp tasks separately, they have the same process title `gulp` which is very hard to differentiate and monitor them. This PR is to solve that issue.


phated:
The CLI embedded in 3.x is frozen.  Can you submit this at https://github.com/gulpjs/gulp-cli (which supports the 3.x and 4 version lines)?

PR #2024, opened 2017-09-29
Handle multiple sources in a task

rajivramv:
An alternate way to handle multiple sources in one task using [`gulp-stream-to-promise`](https://www.npmjs.com/package/gulp-stream-to-promise).


phated:
This is not a recommended pattern.

Issue #71, opened 2013-12-26
How to catch errors without stopping gulp.watch?


dashed:
I'm exploring gulp as a replacement for gruntjs. 

Here is my [gulpfile.js](https://github.com/gobanjs/gojs/blob/cb5be5dc6c843f294807874f97847bf92f688b23/gulpfile.js).

I'm using gulp to watch and compile coffeescript files (with [gulp-coffee](https://github.com/wearefractal/gulp-coffee)). In addition, I'm using gulp.watch to only compile individual files instead of entire targeted globbed paths.

My issue is that if I purposely put incorrect coffeescript syntax, gulp exits:

```
events.js:72
        throw er; // Unhandled 'error' event
< gulp-coffee error here >
< rest of stack trace >
```

How should I catch the error without having it stop gulp.watch? And perhaps re-initiate the default task?

I'm not familiar with streaming in node. I tried try-catch and checked if any callbacks are catching the error to no avail. Appreciate if someone point me in the right direction.

PR #2030, opened 2017-10-17
Docs: Add missing callback calls in code example

phymous:
Gulp pipeline does not work as expected when callback call is missing, e.g. tasks even when `gulp-plumber` is used, immediately stop when errors occur.
This pull request adds the missing callback calls in the code example of the "Writing a plugin" documentation.

### DIFF
--- a/docs/writing-a-plugin/README.md
+++ b/docs/writing-a-plugin/README.md
@@ -141,6 +141,7 @@ module.exports = function() {
         if (file.isStream()) {
             // file.contents is a Stream - https://nodejs.org/api/stream.html
             this.emit('error', new PluginError(PLUGIN_NAME, 'Streams not supported!'));
+            callback();
 
             // or, if you can handle Streams:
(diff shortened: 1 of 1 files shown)


phated:
If we are going to change these docs, it should just use the callback interface completely instead of `emit` + `callback`

phymous:
The `emit` + `callback` combination is already included in this doc [page](https://github.com/gulpjs/gulp/blob/master/docs/writing-a-plugin/dealing-with-streams.md).
Using just `callback(new PluginError(...))` instead does not work with `gulp-plumber` (tasks stop immediately when error is thrown). Invested a lot of time to find this [comment](https://github.com/gulpjs/gulp/issues/71#issuecomment-53942279) and therefore I assumed that the appropriate way is `emit` + `error`.

phated:
Streams are supposed to teardown when they error....

phated:
Closing due to inactivity.

PR #2034, opened 2017-10-31
docs(recipe): Running gulp via cron task

### LINKED ISSUE #2033: '/usr/bin/env: node: No such file or directory' error when running `gulp` via cron job
This is mostly to add some documentation since I am unable to comment on #436.

Basically, if you have a gulp task, say `build`, that you normally run via

    gulp build

And you are trying to add that to a [cron](https://en.wikipedia.org/wiki/Cron) job, you may try something like this: while ssh'd, you run

    which gulp

To get a full path for the global gulp. I get back the path `/usr/local/bin/gulp` (yours might differ).

Anyways, once you have this path, you might decide "run this gulp task every five minutes," which would be setup as the following cron job:

    */5 * * * * cd /your/dir/to/run/in && /usr/local/bin/gulp build

However, you will see in the cron logs that you get this error:

> `/usr/bin/env: node: No such file or directory`

[#436](https://github.com/gulpjs/gulp/issues/436#issuecomment-215435541) directs you to try the following:

> I also did face same issue. This is how I solved it.
> sudo ln -s /usr/bin/nodejs /usr/bin/node
> 
> just in case someone came here looking for solution to this problem.

_However_, that assumes that `/usr/bin/nodejs` is the **actual** path to your node binary.

Really what you want to do is run `which node`, and whatever that path is, is what you want to link to.

In my case, running `which node` shows:

    /usr/local/bin/node

So, first I ended up running

    sudo ln -s /usr/local/bin/node /usr/bin/node

And only after creating that [symbolic link](https://en.wikipedia.org/wiki/Ln_\(Unix\)) did my original cron job execute.

---

## TL;DR:

Run the following if you are having trouble running a gulp task from cron:

    sudo ln -s $(which node) /usr/bin/node

romellem:
The main point of this is to provide documentation
when you get the following error in the cron logs
after trying to run a gulp task:

> /usr/bin/env: node: No such file or directory

Resolves #2033.

### DIFF
--- a/docs/recipes/cron-task.md
+++ b/docs/recipes/cron-task.md
@@ -1,0 +1,25 @@
+# Run gulp task via cron job
+
+While logged in via a user that has privileges to run `gulp`, run the following:
+
+    crontab -e
+    
+to edit your current "[crontab](https://en.wikipedia.org/wiki/Cron)" file.
+
+Typically, within a cron job, you want to run any binary using absolute paths,
+so an initial approach to running `gulp build` every minute might look like:
+
+    * * * * * cd /your/dir/to/run/in && /usr/local/bin/gulp build
+
+However, you might see in the cron logs that you get this error:
+
+> `/usr/bin/env: node: No such file or directory`
+
+To fix this, we need to add a [symbolic link](https://en.wikipedia.org/wiki/Ln_\(Unix\))
+within `/usr/bin` to point to the actual path of our node binary.
+
+Be sure you are logged in as a **sudo** user, and paste in the following command to your terminal:
+
+    sudo ln -s $(which node) /usr/bin/node
+
+Once this link is established, your cron task should run successfully.
--- a/docs/recipes/README.md
+++ b/docs/recipes/README.md
@@ -26,3 +26,4 @@
 * [Run Grunt Tasks from Gulp](run-grunt-tasks-from-gulp.md)
 * [Exports as tasks](exports-as-tasks.md)
 * [Rollup with rollup-stream](rollup-with-rollup-stream.md)
+* [Run gulp task via cron job](cron-task.md)


docs/recipes/cron-task.md
```diff
@@ -0,0 +1,25 @@
...
+While logged in via a user that has privileges to run `gulp`, run the following:
+
+    crontab -e
+    
+to edit your current "[crontab](https://en.wikipedia.org/wiki/Cron)" file.
+
+Typically, within a cron job, you want to run any binary using absolute paths,
+so an initial approach to running `gulp build` every minute might look like:
+
+    * * * * * cd /your/dir/to/run/in && /usr/local/bin/gulp build
```
phated: Why the `* * * * *`?

romellem: Oh that is a `cron` thing. Before the actual command you want to run, you have to specify the schedule for the command. That is, how often a command should run:

```
# ┌───────────── minute (0 - 59)
# │ ┌───────────── hour (0 - 23)
# │ │ ┌───────────── day of month (1 - 31)
# │ │ │ ┌───────────── month (1 - 12)
# │ │ │ │ ┌───────────── day of week (0 - 6) (Sunday to Saturday;
# │ │ │ │ │                                       7 is also Sunday on some systems)
# │ │ │ │ │
# │ │ │ │ │
# * * * * *  command to execute
```

Per [What does * * * * * (five asterisks) in a cron file mean?](https://serverfault.com/q/162388),

> Every minute of every day of every week of every month, that command runs.

I figured I would use that rather some some arbitrary "run once a day at midnight" or something.

phated: Sounds good. Thanks

PR #2037, opened 2017-11-11
Jade was changed to pug

Sylk:


phated:
This was merged without changing the commit to message to our conventional changelog patterns.

PR #2056, opened 2017-12-06
chore: update graceful-fs to v4.0.0

MylesBorins:
Test suite is broken on Node master with graceful-fs 3.x

PR #2057, opened 2017-12-06
chore: remove graceful-fs

MylesBorins:
This is an alternative to #2056 that removes the use of graceful-fs
in the test suite

AFAICT there are no other uses of graceful-fs in gulp proper. We may want to do similar digging into vinyl-fs


phated:
This was completed in f27be05f6d7650ffb3ec3f9375f2d9c50c7010df but thanks for pointing this out!

PR #2064, opened 2017-12-21
Refactoring of code.

Manu-shukla:
Here, I have removed 'var' keyword since we can use only one to define different variables and this will also utilize our memory and load time.


phated:
No thanks.

PR #2066, opened 2017-12-22
Update dependencies

rogierschouten:
Dev-dependencies caused NPM warnings.

Normal dependencies indirectly included an old vinyl version which is becoming a problem because differently shaped objects are travelling the gulp pipelines.


rogierschouten:
By the way, maybe wait with merging this until https://github.com/gulpjs/gulp-util/pull/141 is also merged and published. Then we can also take a new gulp-util with it.

rogierschouten:
I'm unable to run the tests because the package.json scripts do not run on windows.

phated:
I think you missed my comments over there. This looks like it's for master and that's wrong anyway.

PR #2079, opened 2018-01-02
ES2017 async functions

thsmi:
ES2017 standardized the async statement, which makes dealing with promises less painful.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -259,6 +259,8 @@ completion. Tasks are called with a callback parameter to call to signal
 completion. Alternatively, Task can return a stream, a promise, a child process
 or a RxJS observable to signal the end of the task.
 
+In case you use ES2017 (Node 7.6 and up) you can also mark a [function as async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function). This instructs you javascript engine to wrap the function into a promise.
+
 **Warning:** Sync tasks are not supported and your function will never complete
 if the one of the above strategies is not used to signal completion. However,
 thrown errors will be caught by Gulp.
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -315,6 +317,23 @@ $> gulp --tasks
 
 #### Async support
 
+##### ES2017 async function
+
+ES2017 standardized [async functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function). Such function returns by definition always a promise. The javascript engine wraps an async function into a promise and enures that such a promise will be automatically resolved when the async function returns or it will be rejected upon any uncaught exception. 
+
+This makes the explicit done() call required by gulp 4 obsolete and avoids the "Did you forget to signal async completion?" error message. Further more it allows you to write asynchonous code in a synchonous style by using the await statement.
```
demurgos: Thanks for the PR, it's a good idea to mention this.

Gulp (both 3 and 4) support tasks returning Promises. Using async function is more a replacement to returning a Promise than using a `done` callback. I'd just reword this paragraph to say that it provides a lighter syntax over Promises.

phated: Yeah, we handle all sorts of async completion (streams, child process, promises, event emitters, observables & callbacks) so I don't want to make it sound like you can't just return a promise.

phated:
I'd love to get this fixed and merged.  I recently showed off using `async` functions with gulp 4 and everyone loved it.

phated:
I'm thinking about making this a recipe or moving into a separate "async completion" markdown file.

phated:
Now that #2182 is merged, we're documenting this pattern so I'm going to close this.  Thanks for the recommendation @thsmi!

PR #2095, opened 2018-01-11
Docs: Fix problem about the use of gulplog

fangbinwei:
The function(log.info) needs to bind 'log' so that it can emit 'info' event correctly, otherwise, an error 'this'(maybe the instance of Browserify) will be used by the function(log.info).

### DIFF
--- a/docs/recipes/fast-browserify-builds-with-watchify.md
+++ b/docs/recipes/fast-browserify-builds-with-watchify.md
@@ -31,7 +31,7 @@ var b = watchify(browserify(opts));
 
 gulp.task('js', bundle); // so you can run `gulp js` to build the file
 b.on('update', bundle); // on any dep update, runs the bundler
-b.on('log', log.info); // output build logs to terminal
+b.on('log', log.info.bind(log)); // output build logs to terminal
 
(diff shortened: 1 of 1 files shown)


phated:
This might be a bug in gulplog. I'll take a look

fangbinwei:
I think bind `log` is like what `gulp-util` does.
```JavaScript
// gulp-util log.js
var hasGulplog = require('has-gulplog');

module.exports = function(){
  if(hasGulplog()){
    // specifically deferring loading here to keep from registering it globally
    var gulplog = require('gulplog');
    gulplog.info.apply(gulplog, arguments);
  } else {
    // specifically defering loading because it might not be used
    var fancylog = require('fancy-log');
    fancylog.apply(null, arguments);
  }
  return this;
};
```
---
I don't think it is a bug personally. Maybe writing like this is easier to understand in recipe.
```
  b.on('log', function(msg) {
    log.info(msg)
  })
```

phated:
@fangbinwei thanks for pointing this out.  I just published v1.0.1 of glogg which does a lightweight bind to the logger so this doc change shouldn't be needed.

Issue #2099, opened 2018-01-18
README.md Document error, miss 'build' task


RoamIn:
**What were you expecting to happen?**
There is a `build` task
**What actually happened?**
Cann’t find it. (The sample written in ES2015)
![bug](https://user-images.githubusercontent.com/11537595/35077205-e4d8b21e-fc36-11e7-86d8-668a2d867faf.png)

**What version of gulp are you using?**
4.0
**What versions of npm and node are you using?**
It does'nt matter

demurgos:
Instead of:

```javascript
const clean = gulp.series(clean, gulp.parallel(styles, scripts));
gulp.task('clean', clean);
```

It should be:
```javascript
const build = gulp.series(clean, gulp.parallel(styles, scripts));
gulp.task('build', build);
```

PR #2100, opened 2018-01-18
Fix error in ES2015 usage example

demurgos:
The task `clean` was defined twice and the `build` task was missing in
the ES2015 example.
This commit renames the second `clean` task to `build` to match the ES5
example.

Closes gulpjs/gulp#2099


phated:
Thanks @demurgos!

PR #2108, opened 2018-01-28
Docs: Add custom front-matter to "Getting Started"

phated:
cc @somethingnew71 - this is what I was thinking for front-matter.  You can put other stuff docusaurus needs in the comment block


phated:
Done by #2109

PR #2109, opened 2018-01-29
Adds front-matter to all of the docs files

SomethingNew71:
@phated I think I hit them all!


phated:
Awesome, thanks @SomethingNew71!

PR #2129, opened 2018-02-18
Fix recipe - Templating with Swig and YAML front-matter

deslee:
I followed this recipe and was experiencing an issue where livereload was not showing the changes I made to the watched html file. I checked the gulp-swig documentation and disabling caching fixed the issue.


stevelacy:
The recipe does not contain live reload, just bare `gulp-swig` usage - would this not be better just on the `gulp-swig` documentation since the recipe does work as it currently is?

Issue #2137, opened 2018-03-09
Gulp 4.0 Feature request: sync task


buckle2000:
This task will start but will not finish in gulp@4.0:
```
gulp.task('reload', () => console.log('Hi'))
```
I have to change it to
```
gulp.task('reload', async () => console.log('Hi'))
```
- for gulp to work.

Can gulp support synchronic task?
Also, please mention and highlight this behavior explicitly in the API document, thanks.

phated:
No. This was an intentional change

buckle2000:
OK

PR #2138, opened 2018-03-09
Emphasize `asynchronous` in API.md

buckle2000:
#2137


phated:
Thanks but the docs are getting a huge overhaul that explains it better than just saying "async".  I believe the docs for tasks already describes this.

PR #2154, opened 2018-04-14
docs(recipes): Recipe to use browserify with vueify and watchify

Zardoz89:
I noted that there isn't a recipe to build VUE projects with browserify + vueify + babelify and watchify.
I think that this recipe could be useful to newbies to Vue that are trying to use gulp with it.

It's probable that I wrote some bizarre English sentences, so please notice me of anything bad writing on the description of the recipe.


phated:
We are in the process of rewriting all the docs so I'm going to close this.

PR #2172, opened 2018-05-03
add clarification on `.babel.js` suffix

selrond:
see title


phated:
We're working on a complete rewrite of the docs and there's actually going to be a whole page on custom languages.

PR #2182, opened 2018-05-31
Getting Started documentation

phated:
This is the beginning of our full documentation rewrite.  It is meant to be the "Getting Started" section on the upcoming docs website with a target audience of: 
* JS developers that haven't written gulp
* People new to both JS and gulp

It is meant to be read in numerical order (the filenames start with 1,2,3,etc).

Any feedback is very much welcomed and appreciated.

cc @stevelacy @contra @terinjokes @jonschlinkert @doowb @callumacrae @demurgos


phated:
Also posing an open question about whether or not to include documentation/usage of `pump()` for these "intro" docs and examples.  It felt like it would add a bunch of complexity.

docs/getting-started/7-using-plugins.md
```diff
@@ -0,0 +1,113 @@
...
+const through2 = require('through2');
+
+exports.default = function() {
+  return src('src/*.js')
+    // Instead of using gulp-uglify, you can create an inline plugin
+    .pipe(through2.obj(function(file, _, cb) {
+      if (file.isBuffer()) {
+        const code = uglify.minify(file.contents.toString())
+        file.contents = Buffer.from(code)
+      }
```
backflip: @phated, this is probably missing `cb(null, file)` if I'm not mistaken.

backflip:
> Also posing an open question about whether or not to include documentation/usage of pump() for these "intro" docs and examples. It felt like it would add a bunch of complexity.

To me this would mostly depend on whether you consider using `pump()` to be a reasonable default approach. In general, an intro to error handling would definitely make sense to me.

Anyway, great rewrite so far! Would you appreciate help on specific topics now?

docs/getting-started/1-quick-start.md
```diff
@@ -0,0 +1,88 @@
...
+
+This will guide you through giving your project a name, version, description, etc.
+
+### Install the gulp package in your devDependencies
+```sh
+npm install --save-dev gulp
+```
+
+### Verify your gulp versions
+```sh
```
backflip: This would probably need some instructions on what to do with output.

docs/getting-started/3-creating-tasks.md
```diff
@@ -0,0 +1,191 @@
...
+  parallel(
+    cssTranspile,
+    series(jsTranspile, jsBundle)
+  ),
+  parallel(cssMinify, jsMinify),
+  publish
+);
+```
+
+When a composed operation is run, each task will be executed every time it was referenced.  For example, a `clean` task referenced before two different tasks would be run twice and lead to undesired results.  Tasks can be wrapped with the [async-once][async-once] module if this **(not recommended)** pattern is needed.
```
backflip: Is there a recommended pattern for the same problem? Otherwise I think this note isn't very helpful.

docs/getting-started/7-using-plugins.md
phated: Yep, missed that.

docs/getting-started/1-quick-start.md
phated: I think the assumption was that the output should look like the screenshot otherwise you need to do the steps again but maybe that's a poor assumption to make.

docs/getting-started/3-creating-tasks.md
phated: This is a pattern that people migrating from 3.x might encounter so I thought it was worth mentioning that they can use `async-once` but the "right way" to do it would be to restructure your tasks so the task was only specified once - the pattern recommended throughout the document.

I'm open to adjusting the language but I think this is worthwhile enough to keep.

phated:
> To me this would mostly depend on whether you consider using pump() to be a reasonable default approach.

It should be the default going forward (and switched to `stream.pipeline` from node core once 10 becomes LTS and 8 rotates out).

> In general, an intro to error handling would definitely make sense to me.

I like this idea.  I'll discuss with @Janiceilene

phated:
Ping - still looking for y'all to review.

stevelacy:
I did a review of the code vs documentation blocks and did not notice any issues that were not mentioned before. Do we want to show an example of a synchronous task as well as stating that it should not be used?

phated:
Alright all, I've done a final pass and caught some things that I didn't like.  I've also made the fixes suggested so I'm going to get this merged.

PR #2191, opened 2018-06-08
Clarify gulpfile.babel.js example in the README

joeyquarters:
This PR updates the README example to achieve parity with the `require` syntax example and contain error-free code.

**Background**
When following the code example for creating a `gulpfile.babel.js`, I noticed that there was some funkiness in these lines. Specifically, it would actually throw an error due to redefining `const clean`. Hopefully this clears that up and makes the documentation easier to follow for users.

Feedback welcome 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,11 +182,15 @@ function watchFiles() {
 export { watchFiles as watch };
 
 /*
+ * Specify if tasks run in series or parallel using `gulp.series` and `gulp.parallel`
+ */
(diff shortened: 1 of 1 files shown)


stevelacy:
All the examples and documentation are getting a full re-write in this pr: https://github.com/gulpjs/gulp/pull/2182

joeyquarters:
Ah, woops - didn't see that. Will close.

PR #2196, opened 2018-06-24
[WIP] package.json UPDATE WIP FOR TESTING

brodybits:
Change(s) so far:
- migrate from istanbul to nyc

Raising update WIP PR to check Travis CI results (coveralls does not work in on my local system for some reason)


phated:
Heads up, this won't land unless you make it work in node 0.10

brodybits:
Thanks @phated for the heads up. These changes are needed to resolve `npm audit` issues on some `devDependencies` items. Unfortunately the newer packages do not seem to support some older, unsupported versions of Node.js.

I think the best course of action would be to completely drop support for older Node.js versions in a new, major release.

Note that the changes here are really in WIP form, mocha seems to hang in the end for some reason.

phated:
1. devDependencies don't get installed when you `npm install` a package so there's no "security vulnerabilities" with our use of them.
2. npm audit should **not** be auditing sub-dependency devDeps for the above reason - If they are, that's a busted feature.
3. We support old node versions because that's the timeline in which gulp 4 was developed and users shouldn't be forced to use broken node versions to use our project.  It's a build tool...

PR #2204, opened 2018-07-10
Fix typo

AngeloGulina:


stevelacy:
The entire docs are getting rewritten right now #2182

PR #2212, opened 2018-08-26
Update FAQ.md

mrassili:


epancentcev:
ок

pe8ter:
A pattern that I use to avoid both installing Gulp globally and mucking with my `PATH` is to put a reference in the _package.json_:

```json
{
  "scripts": {
    "gulp": "gulp"
  }
}
```

Then when I want to run Gulp:

```bash
$ npm run gulp -- some.task
```

This makes both developing on a team and CI better because it reduces the number of dependencies/configuration needed for all machines involved.

Would this advice be desirable in the documentation?

PR #2226, opened 2018-09-24
API Docs 

phated:
This is a complete rewrite of our API Documentation for gulp 4.  We'd really appreciate some thorough review - as we've stared at these for months.

A few things to note:
* Each API is in it's own document because things were really unwieldy with a single doc previously
* Tables are used extensively to pack a lot of data densely but still be readable.
* API Concepts was created because we had a lot of terminology that kept needing to be defined or linked - I don't believe it is complete but added the things I could come up with.
* Everything should be completely updated for the shipped 4.0 - If not, that's a mistake that should be reported.
* Not all error cases/messages are documented, especially for `src()`/`dest()` because they touch the filesystem and might error in untested ways based on a user's filesytem/configuration.
* We pulled in all the external docs we could hunt down and rewrote them to match our style - so you'll see massive tables of options for `src()` and `watch()` and maybe some others.

I'll comment inline with a few other questions we had on specific parts.

cc @stevelacy @contra @terinjokes @callumacrae @jonschlinkert @doowb @erikkemperman

PR #2230, opened 2018-09-28
Removed not-working link in docs

gbursson:


stevelacy:
All docs are getting a full re-write. https://github.com/gulpjs/gulp/pull/2226

phated:
My gut is that we're going to remove all these during the rewrite.

PR #2277, opened 2018-12-28
Remove gulp@next package

forethoughtde:
Since gulp 4 is released so it is not necessary to install gulp@next


beanilsson:
Perhaps this should be merged? It's April and I'm scratching my head wondering why `npm install --save-dev gulp@next` doesn't work ...

Issue #2153, opened 2018-04-13
gulp.watch not works if watched glob contains Japanese characters folder name


aruku7230:
- gulp: 4.00
- node: 8.9.4
- npm: 5.8.0
- OS: macOS 10.12.6

gulpfile.json
```js
const gulp = require('gulp');
const watchGlob = 'folder/**/*';

gulp.task('test', () => {
    gulp.watch(watchGlob, (done) => { done() })
            .on('change', (path) => {
                console.log(`File ${path} changed.`);
            });
});
```
run test task and change `folder/foo` file
result output is: 
> Starting 'test'...
> File folder/foo changed.
> Starting '<anonymous>'...
> Finished '<anonymous>' after 1.45 ms

change watchGlob
```js
const watchGlob = 'フォルダ/**/*';
```
run test task and change `フォルダ/foo` file. 
result output is: 
> Starting 'test'...

contra:
Can you test it with https://github.com/paulmillr/chokidar directly and see if you can reproduce it? If so the issue goes there, if not we can look into it here.

seminaoki:
Reproduce it use chokidar. It's a issue with chokidar.

```javascript
const watchGlob = 'フォルダ/**/*';

gulp.task('test', () => {
    const chokidar = require('chokidar');
    chokidar.watch(watchGlob, (done) => { done() })
            .on('change', (path) => {
                console.log(`File ${path} changed.`);
            });
});
```

sttk:
This issue is due to `micromatch` which is used in `chokidar`.

MasatoMakino:
I think it related to NFD filename on macOSX.
NFD's "ダ" is "タ" + " " ".
NFD's "ダ" chrcode deffer from NFC's "ダ" chrcode.

phated:
We've just added a Bug Bounty of $50 on this issue if anyone wants to solve it.

pe8ter:
I think I have the issue narrowed down to a `RegExp` in the `nanomatch` package. This issue is not restricted to Japanese, but to any directory that doesn't end in `0-9`, `a-z`, `A-Z`, or `_`. I opened https://github.com/micromatch/nanomatch/issues/18.

jonschlinkert:
> but to any directory that doesn't end in `0-9`, `a-z`, `A-Z`, or `_`.

The regex was written to only match valid path characters, but this clearly was shortsighted and insensitive to non-English languages.  I will get that fixed.

[...]

PR #2299, opened 2019-03-08
Add Azure Pipelines CI

damccorm:
**The Change**

This PR adds Azure Pipelines to perform CI builds and PR builds. Pipelines offers several key advantages:

1) Single definition for all platforms (Linux/Windows/Mac) reduces the amount of time spent maintaining CI.
2) Faster overall build times because of free parallelism - Pipelines offers 10 free concurrent jobs and unlimited build minutes for all OSS projects. Plus, for big projects like this one, Microsoft is happy to raise the concurrency limit, so we could have 15 or 20 parallel jobs. This should cut max build times to 4 or 4.5 minutes on Pipelines (it seems to vary a lot for Appveyor, but I've seen anything from 6 - 10 minutes in previous builds).
3) Free MacOS builds - right now we're not building on MacOS at all, but [some issues appear only on that OS](https://github.com/gulpjs/gulp/issues/2153).

**Disclaimer**

Full disclosure, I'm an engineer on the Pipelines team, but I also think Pipelines provides some pretty significant advantages here.

**Additional Info**

You can see that this works in [my pipeline](https://dev.azure.com/gulpci/gulp/_build/results?buildId=84).

image

With test reporting!

image


phated:
This is actually something I've been considering.

I won't have any time today to dig in but I'll try to find time soon.

Can you remove the package-lock file that was generated?  I haven't yet added the npmrc file that stops them from being generated.

damccorm:
Whoops, didn't realized I'd added that, taken care of 😄

.azure-pipelines-steps.yml
```diff
@@ -0,0 +1,27 @@
...
+  displayName: Run coveralls
+  condition: eq(variables['run_coveralls'], true)
+
+- script: npm test
+  displayName: Run tests
+  condition: ne(variables['run_coveralls'], true)
+
+- task: PublishTestResults@2
+  inputs:
+    testResultsFiles: '**/xunit.xml'
```
phated: So azure pipelines only speaks xunit?

.azure-pipelines-steps.yml
```diff
@@ -0,0 +1,27 @@
+steps:
+- script: npm i -g npm@$(npm_version)
+  displayName: Use legacy npm version $(npm_version)
```
phated: The node versions and the npm versions are aligned?  In travis and appveyor, when we "checkout" an node version, we get the distributed npm version

.azure-pipelines-steps.yml
```diff
@@ -0,0 +1,27 @@
...
+  condition: ne(variables['npm_version'], '')
+
+- task: NodeTool@0
+  inputs:
+    versionSpec: '$(node_version)'
+  displayName: Use Node $(node_version)
```
phated: What does this do?  Just print the node version?

.azure-pipelines-steps.yml
```diff
@@ -0,0 +1,27 @@
...
+
+- script: npm install
+  displayName: npm install
+
+- script: npm run coveralls
+  env:
+    COVERALLS_REPO_TOKEN: $(COVERALLS_REPO_TOKEN_SECRET)
```
phated: Is this needed?  Currently coveralls just knows that the coverage is being pushed from our public repo.

.azure-pipelines.yml
```diff
@@ -0,0 +1,66 @@
...
+          npm_version: 2
+    steps:
+      - template: .azure-pipelines-steps.yml
+
+  - job: Test_MacOS
+    displayName: Run Tests on MacOS
+    pool:
+      vmImage: macos-10.13
+    strategy:
+      matrix:
```
phated: Should we be testing all versions of node on Mac?  If it is fast enough, we probably should because I have a lot of Mac-specific fixes in vinyl-fs

package.json
```diff
@@ -43,6 +43,7 @@
     "istanbul-coveralls": "^1.0.3",
     "mkdirp": "^0.5.1",
     "mocha": "^3.0.0",
+    "mocha-multi-reporters": "^1.1.7",
```
phated: I'd prefer not to add another devDep if possible

.azure-pipelines-steps.yml
damccorm: With this input it does, but it can also use JUnit, NUnit, VSTest, xUnit, or cTest, we'd just need to change testResultsFiles to something like `**/junit.xml`. I believe if we want we could also do something like `**/TEST-**.xml`, this input really just works as a pattern matcher I think.

I used xunit since that matches the custom reporter I added (more on that below).

.azure-pipelines-steps.yml
damccorm: We don't do that by default yet, I think I agree that we should have this as an option going forward though. If we want to make sure we have exact alignment we can run this step for every job for now.

I'm going to be working on some of our node tooling over the next couple of weeks, I'll try to make this better.

.azure-pipelines-steps.yml
damccorm: It installs the specified version of node. Its similar to saying `nodejs: {node_version}` in Travis, we're just drawing the node version from the variable we defined in `.azure-pipelines.yml`

.azure-pipelines-steps.yml
damccorm: Yes, unfortunately Coveralls isn't as tightly integrated with Pipelines yet. I think that should change going forward, but for now we need to pass in a token to authenticate. You can get that from here:

image

I'd also recommend keeping it a [secret variable](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables) to make sure it doesn't get leaked. That's how I have it set up in my configuration as well.

.azure-pipelines.yml
damccorm: Yeah. Not sure why I wasn't just doing that to begin with. Updated appropriately

package.json
damccorm: That's probably fine. The advantage of this is that it allows us to use both the default reporter which prints live output and also creates a test results file. This allows us to produce the nicer test result reporting.

One workaround we could use here is to install this as a step in our ci process. The downside is that its a little less clean, we'd have to add that step and explicitly call `mocha --reporter mocha-multi-reporters --async-only` instead of just `npm test` in the testing step. We can also just not report results for that step.

Thoughts?

.azure-pipelines-steps.yml
damccorm: Actually, did a little more digging on this. It turns out in the general case, we behave like we should (installing the appropriate version of npm). The only exception to this is on older versions of windows. This is because node doesn't automatically package them in their distribution in the same way. I still think it should be improved, but it only affects node<5 on Windows I think.

In the meantime, the current legacy install step should be sufficient to make sure the packages match the distribution so we don't need to make any immediate changes here.

phated: @damccorm thanks for looking into that!  I actually remember the "bad old days" when npm didn't ship with the Windows distribution.  Once we start dropping old versions (a ways off), we'll be able to reduce these phases.

.azure-pipelines-steps.yml
phated: Ah, okay.  So the `NodeTool@0` task is installing it.  That makes sense.

.azure-pipelines-steps.yml
phated: Alright.  Yeah, it'd be nice to have deeper coveralls integration.

Do we have all these environment variables that coveralls says we need:
```
CI_NAME
CI_BUILD_NUMBER
CI_BUILD_URL
CI_BRANCH
CI_PULL_REQUEST (optional)
```

.azure-pipelines.yml
phated: Thanks!

.azure-pipelines-steps.yml
phated: @damccorm It looks like mocha (even the version we use) [ships with an XUnit reporter](https://mochajs.org/#xunit).  Can we just use that?  We can have an `azure` script that does `mocha --async-only --reporter xunit`

That way we can drop the extra reporter.

package.json
phated: I mentioned it above but it looks like mocha ships with an XUnit reporter, so I think we just need to make this a separate npm script and add another phase.  Would that make sense?

phated:
@damccorm Just a few more things I followed up on.  Also, do you have any guidance on how I signup and get this configured to work under a gulpjs organization for Azure Pipelines?

phated:
Oh, also - does Azure Pipelines support a `.ci/` directory?  I'd like to start moving things there if possible.

package.json
damccorm: I responded in the above thread, lets consolidate and continue conversation there 😄

.azure-pipelines-steps.yml
damccorm: That works, the only downside is we don't get great output in the console.

What we get with multi-reporter:

image

What we get with xunit reporter:

image

I have this change in PR, if you're comfortable with that tradeoff I'll merge just let me know.

.azure-pipelines-steps.yml
damccorm: I'm guessing you're drawing that from [here](https://docs.coveralls.io/supported-ci-services), I believe that's just for Ruby, though I'm not 100% sure - it you look under the "configuration" heading it says "This configuration guide is specifically for the Ruby language. Other language support can be found in the individual READMEs.".

I used the instructions in the [node-coveralls' usage section](https://github.com/nickmerwin/node-coveralls#usage) which say we need COVERALLS_SERVICE_NAME and COVERALLS_REPO_TOKEN. The reason COVERALLS_REPO_TOKEN is specifically called out is that its a secret variable, and secrets aren't scoped into tasks by default (to avoid them leaking to malicious or careless custom tasks). COVERALLS_SERVICE_NAME is also stored in the env though.

image

damccorm: As a side note, I'm adding coveralls integration to our backlog, definitely agree this would be a good thing to have.

damccorm:
> Oh, also - does Azure Pipelines support a .ci/ directory? I'd like to start moving things there if possible.

Sure, I moved it in there in my PR.

> do you have any guidance on how I signup and get this configured to work under a gulpjs organization for Azure Pipelines?

The easiest way to get this done is probably to install the [GitHub app](https://github.com/marketplace/azure-pipelines) after you've merged in this PR. That will walk you through the process of creating an organization and project. From there you can follow the getting started docs [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/get-started-yaml?view=azure-devops&viewFallbackFrom=vsts).

One thing to call out is that since we moved the yaml to the `.ci` folder it won't be automatically discovered immediately when you create a build pipeline. That just means you'll want to select the `Existing Azure Pipelines YAML file` configuration and set `/.ci/.azure-pipelines.yml` as your path.

Let me know if anything is unclear as well, happy to help however I can!

damccorm:
One more thing - I noticed that both the organization names `gulp` and `gulpjs` are taken. I'm happy to transfer `gulpci` to you as an org name if you can't find a better one. It also looks like `gulpjs` is currently owned by team@gulpjs.com, so maybe your team already has one?

Regardless, installing the GitHub app is still the best way to get going.

.azure-pipelines-steps.yml
damccorm: Oh actually, on closer inspection, this doesn't get written to a file which is necessary for us to report results. We can pipe it out to a file, but that would mean we would get no output at all on the console.

phated: We can use the `-O output=test.xunit` flag to specify the output file.  I was thinking we could just run the test suites as separate phases, one with the spec reporter and one with the xunit reporter.

.azure-pipelines-steps.yml
phated: Ah okay!  Thanks for adding it to the backlog 😍

phated:
@damccorm yeah, I did set one up a long time ago but I gave up when I ran into tons of XML.  Is the YAML stuff new?  Can I use the existing team if I install it as a GitHub App?

.azure-pipelines-steps.yml
damccorm: Makes sense to me! Updated accordingly

damccorm:
> Is the YAML stuff new?

Yeah, it was added a little more than half a year ago as the default experience. It was part of a larger push (which is still ongoing) to offer a better OSS experience in particular.

> Can I use the existing team if I install it as a GitHub App?

Yep! It should give you an opportunity to sign into an existing organization.

phated:
Sorry for the delay.  This is looking great!

phated:
@damccorm I'm looking at the docs for secure variables and it seems to indicate there should be a 🔒  icon in the variables panel but I don't have that.  How do I secure this coveralls token?  Also, by using the coveralls token as a secure token, it says it won't be available to PRs - won't that break our coveralls Status API integration for PRs?

damccorm:
> there should be a 🔒 icon in the variables panel but I don't have that

Are you looking at this page or somewhere else? This is what I see (with the secure checkbox column highlighted):

image

If not could you provide a little more info on what you're looking at? To get to that page, you can follow these steps:

Edit pipeline
image

Choose `variables` from the dropdown here:
image

> Also, by using the coveralls token as a secure token, it says it won't be available to PRs - won't that break our coveralls Status API integration for PRs?

I missed that you were doing this. The answer is yes by default, but you can configure this behavior from the same menu (under the triggers tab)

image

If you do this, I would be careful though. The reason that secrets aren't enabled on forked builds is that someone could change a script that a secret is mapped into and do something malicious (e.g. send themselves the token). They can't change the yaml that gets executed though (or at least it won't get executed until merged), so a change you could make to ensure this doesn't happen is instead of calling `npm run coveralls` running the actual script (`npm run cover && istanbul-coveralls`). I'll add a PR to do this.

phated:
@damccorm thanks for following up.  I have totally different variables view than you.  Is it possible that I'm on a newer/older version of the UI because I signed up so long ago?  I also noticed that other stuff in the documentation didn't match my UI.

phated:
@damccorm this is what I'm seeing for my "variables" screen
Screen Shot 2019-03-20 at 2 12 36 PM

damccorm:
Huh, that's weird, I haven't seen that view before (I joined Pipelines relatively recently though, so there's a chance it predates me). One thing that you could do is check to make sure the new UI didn't somehow get disabled in the preview features. For me you can access this here, hopefully this works for you as well:

image

damccorm:
I'll also reach out internally and get back to you on this one, I don't know exactly why its behaving this way

phated:
@damccorm there was some stuff disabled in the "Preview Features" but I've now enabled everything and still don't have that UI.  If it gives any indicators, the URL I'm accessing is https://dev.azure.com/gulpjs/gulp/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=1&view=Tab_Variables

phated:
@damccorm LMAO - that UI comes from having the sidebar.  I just closed it and noticed more to the UI.  I never even tried to scroll that pane because there was no indicator that it had overflowed the panel.  Any chance you could add that to your issue tracker????

damccorm:
Yikes. Yeah, that's not good, we'll address that.

Issue #2164, opened 2018-04-25
[FEEDBACK REQUESTED]


Janiceilene:
We're doing a full rewrite of the docs. We've noticed that recipes is pretty unruly and the quality varies a lot.  We need to trim, edit, and add to them.  

We'd love to get some feedback on which recipes you use, which you don't touch, and what new ones you'd like to see!

demurgos:
Hi,
I just wanted to say that the [Delete files and folders](https://github.com/gulpjs/gulp/blob/master/docs/recipes/delete-files-folder.md) recipe helped me when I started with Gulp. Deleting files is a common problem and the fact that you don't need a special plugin but can just use normal libraries is a good thing to know.

I did not use other recipes. Thanks for helping with the documentation.

polybuildr:
I loved the handling-the-delete-event-on-watch recipe: it helped me get rid of many of the problems I was facing with my gulp setup previously.

buksy90:
Hello, I would love to see classic documentation with functions signatures, explained inputs/outputs, listed input flags and examples on use :). I believe most of this could be autogenerated (except the examples).

garza:
I relied a lot on the server-with-livereload example to get our web project up and running

dominic-p:
I would personally love to see a recipe for integrating with webpack. I hacked something together, but I have the feeling it could be a lot better.

jaireina:
What I do most of the time:

- SASS + Source maps
- Pug incremental builds
- ESLINT

Something that I remember took me some time to sort out:
- React / JSX / BABEL
- Being able to use ES6 code in the gulp file itself.

bennyn:
Here is my recipe on how to use **Jasmine 3.2 with gulp 4**:

```javascript
const chalk = require('chalk');
const gulp = require('gulp');
const Jasmine = require('jasmine');
const path = require('path');

const gutil = {
  log: require('fancy-log'),
};

gulp.task('test:node', (done) => {
  const jasmine = new Jasmine();
  jasmine.loadConfigFile(path.join(process.cwd(), 'spec', 'support', 'jasmine.json'));
  jasmine.configureDefaultReporter({
    showColors: true
  });

  jasmine.onComplete((passed) => {
    if (passed) {
      gutil.log(chalk`All specs have {green passed}.`);
      done();
      process.exit(0);
    }
    else {
      process.exit(1);
    }
  });

  jasmine.execute();
});
```

[...]

PR #2305, opened 2019-03-19
Updating recipes to align with new docs

S1SYPHOS:
Let me know if migrating over some / all of the recipes is something you are interested in.

### DIFF
--- a/docs/recipes/using-external-config-file.md
+++ b/docs/recipes/using-external-config-file.md
@@ -1,6 +1,6 @@
 # Using external config file
 
-Beneficial because it's keeping tasks DRY and config.json can be used by another task runner, like `grunt`.
+Beneficial because it keeps tasks DRY and `config.json` can be used by another task runner, like `grunt`.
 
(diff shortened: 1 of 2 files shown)


phated:
The recipes really need to be deleted.  They are a massive disappointment.  #2164 is our epic that outlines that they all need to go away or be completely rewritten.

S1SYPHOS:
True. I wanted to kickoff a complete rewrite, but since it's unclear which examples should stay, it's all or nothing I guess ..

phated:
I'd be more than willing to give technical feedback on a full rewrite.  I might even be able to rope @janiceilene in for general content editing.

Do you want to take that on?  I think it'd probably be best to rewrite them all, then during review, I can decide if I don't like the recipe existing and then we can delete it.

S1SYPHOS:
I'll look into it and whip something up.

phated:
@s1syphos looks like you are no longer going to contribute these?

PR #2306, opened 2019-03-20
Explicitly run coveralls step

damccorm:
This will allow the token to be passed on PRs in without concern of it being stolen by a changed script. Without this, an attacker could change the `npm run coveralls` command in package.json in their PR so that it does something bad (e.g. sends them the token).

See https://github.com/gulpjs/gulp/pull/2299#issuecomment-475003985

### DIFF
--- a/.ci/.azure-pipelines-steps.yml
+++ b/.ci/.azure-pipelines-steps.yml
@@ -1,0 +1,30 @@
+steps:
+- script: npm i -g npm@$(npm_version)
+  displayName: Use legacy npm version $(npm_version)
+  condition: ne(variables['npm_version'], '')
+
+- task: NodeTool@0
+  inputs:
+    versionSpec: '$(node_version)'
+  displayName: Use Node $(node_version)
+
+- script: npm install
+  displayName: npm install
+
+- script: npm run coveralls
+  env:
+    COVERALLS_REPO_TOKEN: $(COVERALLS_REPO_TOKEN_SECRET)
+  displayName: Run coveralls
+  condition: eq(variables['run_coveralls'], true)
+
+- script: npm test
+  displayName: Run tests
+  condition: ne(variables['run_coveralls'], true)
+
+- script: npm run azure-pipelines
+  displayName: Write tests to xml
+
+- task: PublishTestResults@2
+  inputs:
+    testResultsFiles: '**/test.xunit'
+  condition: succeededOrFailed()
--- a/.ci/.azure-pipelines-steps.yml
+++ b/.ci/.azure-pipelines-steps.yml
@@ -11,7 +11,7 @@ steps:
 - script: npm install
   displayName: npm install
 
-- script: npm run cover && istanbul-coveralls
+- script: istanbul cover _mocha --report lcovonly && istanbul-coveralls
   env:
     COVERALLS_REPO_TOKEN: $(COVERALLS_REPO_TOKEN_SECRET)
   displayName: Run coveralls
--- a/.ci/.azure-pipelines-steps.yml
+++ b/.ci/.azure-pipelines-steps.yml
@@ -11,7 +11,7 @@ steps:
 - script: npm install
   displayName: npm install
 
-- script: npm run coveralls
+- script: npm run cover && istanbul-coveralls
   env:
     COVERALLS_REPO_TOKEN: $(COVERALLS_REPO_TOKEN_SECRET)
   displayName: Run coveralls
--- a/.ci/.azure-pipelines.yml
+++ b/.ci/.azure-pipelines.yml
@@ -1,0 +1,72 @@
+trigger:
+- master
+- releases/*
+
+jobs:
+  - job: Test_Linux
+    displayName: Run Tests on Linux
+    pool:
+      vmImage: "Ubuntu 16.04"
+    variables:
+      run_coveralls: true
+    strategy:
+      matrix:
+        Node_v10:
+          node_version: 10
+        Node_v8:
+          node_version: 8
+        Node_v6:
+          node_version: 6
+        Node_v4:
+          node_version: 4
+        Node_v0_12:
+          node_version: 0.12
+        Node_v0_10:
+          node_version: 0.10
+    steps:
+      - template: .azure-pipelines-steps.yml
+
+  - job: Test_Windows
+    displayName: Run Tests on Windows
+    pool:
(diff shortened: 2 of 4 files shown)


.ci/.azure-pipelines-steps.yml
```diff
@@ -11,7 +11,7 @@ steps:
 - script: npm install
   displayName: npm install
 
-- script: npm run coveralls
+- script: npm run cover && istanbul-coveralls
```
phated: @damccorm With this change, couldn't they change the `npm run cover` command and have access?

damccorm: Yes 🤦‍♂️ Updated appropriately, sorry about that

.ci/.azure-pipelines-steps.yml
```diff
@@ -11,7 +11,7 @@ steps:
 - script: npm install
   displayName: npm install
 
-- script: npm run coveralls
+- script: istanbul cover _mocha --report lcovonly && istanbul-coveralls
```
phated: I know I'm being a little paranoid here but what would happen if someone changed the `package.json` to point `istanbul` or `istanbul-coveralls` at some git repository?

damccorm: Can someone do that? My understanding of how npm works is that it makes that sort of behavior impossible. They'd have to change the dependency name in `package.json` at which point it would no longer be added to the path as `Istanbul` or `Istanbul-coveralls`, it would get added as something else.

phated: I'm allowed to do something like:
```json
{
    "devDependencies": {
        "istanbul": "github:phated/malicious-package"
    }
}
```
And as long as "malicious-package" contained an `istanbul` bin, it'd run.

phated: Maybe this step needs to forcefully install the packages from npm.  We can probably achieve that by using `npx --ignore-existing istanbul && npx --ignore-existing istanbul-coveralls` - Thoughts?

lirantal:
@damccorm this issue is referring only to azure pipelines integration, right? because Travis CI doesn't provide token access to forks.

.ci/.azure-pipelines-steps.yml
damccorm: Ah, I see what you mean now - I guess that could be an issue. I think that solution should work in theory, I'm having a bit of trouble getting it to work though because it looks like npx doesn't work correctly on lower versions of npm or Node - see [build here](https://dev.azure.com/gulpci/gulp/_build/results?buildId=95).

We can keep looking for a solution here, to be honest I'm not sure how necessary this is though. I don't think there's anything a malicious actor could do with this token outside of messing with your code coverage results - the work to do that seems to far outweighs the benefits, especially since you can just regen the token and delete the bad builds. In the codecov docs they don't even recommend keeping it as a secret variable even though this would affect all CI's outside of Travis and I think most repos that use this without Travis just keep it as plaintext.

If you want I can keep trying to tweak npx though, I'm happy to continue to work on this if we think its worthwhile. Just don't want to overcomplicate things if we don't need to.

damccorm:
I'm not sure. I think they might actually just automatically inject the COVERALLS_REPO_TOKEN into their builds in which case it would be an issue, but they might have a more tightly integrated approach. I don't know exactly how this works on their end.

phated:
@damccorm I'm having a really hard time getting coveralls to work correctly.  They need a unique ID that represents the entire job and then individual job IDs to group under the top-level job.  I don't see anything like that in https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml

damccorm:
I think we can definitely use `Build.BuildNumber` for the unique Id across the build. For the individual job IDs, could we use the version of node as the id? That should be unique for each coveralls run per build.

phated:
@damccorm I was able to get this to work with these environment variables: https://github.com/gulpjs/gulp/blob/b2c6c7/.ci/.azure-pipelines-steps.yml#L19-L30 and this extra step https://github.com/gulpjs/gulp/blob/b2c6c7e4c6cc85e6f7579b3c32081590d9517728/.ci/.azure-pipelines.yml#L72-L84 - which produces: https://coveralls.io/builds/22365839

Any ways to clean that up would be greatly appreciated!

phated:
@damccorm it seems like we can't actually enable Azure Pipelines to run on Forks because of this little snippet I found in the docs: 

>By default with GitHub pipelines, secrets associated with your build pipeline are not made available to pull request builds of forks. These secrets are enabled by default with GitHub Enterprise Server pipelines. Secrets include:

> * A security token with access to your GitHub repository.
> * These items, if your build uses them:
>     * Service connection credentials
>     * Files from the secure files library
>     * Build variables marked secret

That seems to indicate that "Make secrets available to builds of forks" is an all-or-nothing option.  I'd either have to make the Coveralls token non-secret or expose our repo to attacks using it's security token.  I really wish I could decide which secrets to allow to exposed forks.  It'd also be really helpful if I could reject Azure Pipelines running if certain files were changed (like my package.json, etc) - it doesn't solve everything but it helps a lot.

I'm going to close this for now and just concede that we need to run Travis, AppVeyor & Azure Pipelines side-by-side until Azure fits all our use cases better.

damccorm:
@phated where are you pulling that from? It seems pretty misleading to me and I'd love to clean the doc up.

To be clear, secrets are not scoped into tasks by default, regardless of whether secrets are enabled on forks. So unless you explicitly scope the security token into a task (which I definitely wouldn't recommend), forks will have no way of accessing that token.

So if you have the following YAML definition:
```
- task: task1
  env:
    MY_SECRET_1: $(MY_SECRET_1)

- task: task2
  env:
    MY_SECRET_2: $(MY_SECRET_2)
```

task1 has access to `MY_SECRET_1` but not `MY_SECRET_2` and vice versa. So unless you explicitly scope the security token with access to your repo in, no task will be able to use it.

Not sure if that changes how you view this or not, but just wanted to be clear on that. Several Microsoft repos enable secrets on forks, and that would be a big-time safety issue if people could attack the repos through that.

phated:
@damccorm oh, that's good to know.  I found those docs at https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops#validate-contributions-from-forks

For the time being, I just enabled the `/azp` command thing that I can trigger after reviewing code.

PR #2310, opened 2019-03-25
Consider adding Azure Pipelines badge

damccorm:
Follow up to #2299. Would probably be good to have if we're deciding to keep Pipelines around.


phated:
I totally meant to do this but got bogged down by the Coveralls stuff and totally forgot.  Thanks!

PR #2312, opened 2019-03-27
To be more explicit regarding transpilation

### LINKED ISSUE #2311: Module resolution for gulpfile.ts directory
The [docs](https://github.com/gulpjs/gulp/blob/master/docs/getting-started/2-javascript-and-gulpfiles.md) describe module resolution to a folder named gulpfile.js.  Somewhere there should be a mention that this will not work with a folder gulpfile.ts when using gulp-typescript even if it is a given that node module resolution works the same here.

tonyhallett:
closes #2311

### DIFF
--- a/docs/getting-started/2-javascript-and-gulpfiles.md
+++ b/docs/getting-started/2-javascript-and-gulpfiles.md
@@ -15,7 +15,7 @@ A gulpfile is a file in your project directory titled `gulpfile.js` (or capitali
 
 ## Transpilation
 
-You can write a gulpfile using a language that requires transpilation, like TypeScript or Babel, by changing the extension on your `gulpfile.js` to indicate the language and install the matching transpiler module.
+You can write a gulpfile using a language that requires transpilation, like TypeScript or Babel, by changing the extension on your `gulpfile.js` file or folder ( see Splitting a gulpfile ) to indicate the language and install the matching transpiler module.
 
 * For TypeScript, rename to `gulpfile.ts` and install the [ts-node][ts-node-module] module.
 * For Babel, rename to `gulpfile.babel.js` and install the [@babel/register][babel-register-module] module.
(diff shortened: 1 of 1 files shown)


phated:
@tonyhallett do you have to name both things?  I would guess the actual files would be the only thing that needs the extension change.

tonyhallett:
@phated Ok.  Can I just create another commit on the patch branch ?

phated:
Yes, that is fine.

tonyhallett:
Ok will do

docs/getting-started/2-javascript-and-gulpfiles.md
```diff
@@ -28,7 +28,7 @@ Many users start by adding all logic to a gulpfile. If it ever grows too big, it
 
 Each task can be split into its own file, then imported into your gulpfile for composition. Not only does this keep things organized, but it allows you to test each task independently or vary composition based on conditions.
 
-Node's module resolution allows you to replace your `gulpfile.js` file with a directory named `gulpfile.js` that contains an `index.js` file which is treated as a `gulpfile.js`. This directory could then contain your individual modules for tasks.
+Node's module resolution allows you to replace your `gulpfile.js` file with a directory named `gulpfile.js` that contains an `index.js` file which is treated as a `gulpfile.js`. This directory could then contain your individual modules for tasks.  If you are using a transpiler then name the folder and file accordingly.
```
janiceilene:
```suggestion
Node's module resolution allows you to replace your `gulpfile.js` file with a directory named `gulpfile.js` that contains an `index.js` file which is treated as a `gulpfile.js`. This directory could then contain your individual modules for tasks.  If you are using a transpiler, name the folder and file accordingly.
```

Tiny change!

phated: Thanks @janiceilene! @tonyhallett it doesn't seem like you allowed gulp contributors to make changes so you'll need to accept this suggestion then I can merge.

phated:
Thanks @tonyhallett!

PR #2326, opened 2019-04-26
Update 6-explaining-globs.md

ralomach:
I was reading the documentation and maybe found 2 typos, refering to the "scripts" folder as "script", when all the examples have "scripts".


phated:
Good catch!  Thanks 😸

Issue #2355, opened 2019-07-19
Upgrade nyc to a version not affected by https://nodesecurity.io/advisories/577   


apupier:
even if it is a dev dependency it will make the build cleaner and easier to detect potential security issue with dependency

phated:
This is not possible due to our build matrix. It is a devDependency so it doesn't affect anything. If you'd like to send a PR to disable `npm audit` on install to clean up the builds, we're happy to accept that.

PR #2356, opened 2019-07-19
Upgrade nyc to 14.1.1 #2355

apupier:
fixing vulnerabilities (16 low, 19 high) even if it was affecting only
dev dependencies, it helps to see clearer

Signed-off-by: Aurélien Pupier <apupier@redhat.com>


phated:
Gulp 4 supports node 0.10. Making this change would be a breaking change and will be done once gulp reaches a point where it makes sense to release a major semver bump.

Issue #2357, opened 2019-07-19
Upgrade mocha


apupier:
to a version fixing https://www.npmjs.com/advisories/146

phated:
This is not possible due to our build matrix. It is a devDependency so it doesn't affect anything.

PR #2358, opened 2019-07-19
Upgrade Mocha from 3.0.0 to 6.2.0 #2357

apupier:
fixing 1 critical security issue reported by npm


phated:
Gulp 4 supports node 0.10. Making this change would be a breaking change and will be done once gulp reaches a point where it makes sense to release a major semver bump.

PR #2365, opened 2019-08-13
fix broken doc code

mimoduo:
Hey gulpjs,

First off thank you for the awesome build tool!

I'm in the process of rewriting my front end gulpfile and found that the example for uglifying js passed an object into a Buffer that accepts a string. This commit fixes that reference.

Sincerely,
Bryan Stoner


phated:
Thanks!

PR #2382, opened 2019-09-30
Typo in last line of usage example

rninne:
After changing the stem to foo and the ext to .js, the full path should be `/specs/foo.js`


phated:
Thanks!

PR #2386, opened 2019-10-12
Security improment using npm audit fix -f

ralic:

PR #2388, opened 2019-10-22
Security improvement

ralic:
Original Pull Request : 
https://github.com/gulpjs/gulp/pull/2386

Please reconsider this pull request again because :

1 ) All test passed .
https://travis-ci.org/SaferNodeJS/gulp
https://travis-ci.org/SaferNodeJS/gulp/builds

Security improvement reported as following
Before
```
added 675 packages from 696 contributors and audited 7707 packages in 65.48s
found 47 vulnerabilities (17 low, 7 moderate, 21 high, 2 critical)
  run `npm audit fix` to fix them, or `npm audit` for details
```
After Fixing
```
added 635 packages from 822 contributors and audited 7354 packages in 41.695s
found 0 vulnerabilities
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,6 @@ node_js:
   - '12'
   - '10'
   - '8'
-  - '6'
-  - '4'
-  - '0.12'
-  - '0.10'
 install:
   - npm install -g grunt-cli
   - npm install
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,12 @@ node_js:
   - '4'
   - '0.12'
   - '0.10'
+install:
+  - npm install -g grunt-cli
+  - npm install
+  - npm run test
 after_script:
   - npm run coveralls
+notifications:
+  email:
+    on_success: never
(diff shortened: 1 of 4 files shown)


demurgos:
@ralic 
This is a breaking change. Even if the Node maintainers and some plugins dropped support for older Node versions, Gulp still supports them (yes, even Node 0.10).

If you want to fix these vulnerabilities, the first step is to open an issue to discuss the supported Node versions.

ralic:
The majority of users using gulp shall be developers. Node.js could be updated either through yum or apt on Linux or download directly from node.js website in other systems. Getting an LTS version of Node.js shall not be hard for a developer, instead, getting v0.12 node.js may take more chances for the software being developed getting hacked.   I think this is why the npm audit is getting developed: to provide safety. 

Perhaps, it might be to like Android System, less than 1 % would use the latest SDK. While in the opposite, Apple request their developer to support their new SDK. It is a sad story to see google's new Android SDK is deprecated.

phated:
No. I've already detailed this in many other issues and I'm sick of explaining it. Use the search.

ralic:
What if the lower version node.js test failure gets resolved?

I mean that I can try to fix tests for these new dependencies to be more compatible for the older Node.js version.

PR #2389, opened 2019-10-24
Update running-shell-commands.md recipe

azerafati:
Adding example of `spawn` and how to print output to console.


phated:
We're temporarily halting recipe additions or updates. Sorry.

PR #2396, opened 2019-11-26
Added stream error handling

stgogm:
Some errors, like unexpected tokens, cause the Gulp task to crash and this will avoid that.


phated:
You shouldn't be avoiding failures. The watcher handles failures in gulp 4.

PR #2406, opened 2020-01-11
Correct typo in rollup example

aaronhooper:
Rollup is imported by name, so it should be called like `rollup()`.


phated:
Good catch! Thanks 😄

PR #2428, opened 2020-04-02
Fix for code example in documentation (Getting Started/Async Completion)

chinhdo:
The code example on the [Async Completion](https://gulpjs.com/docs/en/getting-started/async-completion) page was meant to log the version from package.json to the console, but it was logging out "undefined".

Output of previous code example:
image

Output of fixed code example:
image


phated:
Thanks!

PR #2466, opened 2020-07-15
Update rollup recipe

hirusi:
Just a couple of deprecated properties. Additionally, specifying format is mandatory now or rollup complains.

### DIFF
--- a/docs/recipes/rollup-with-rollup-stream.md
+++ b/docs/recipes/rollup-with-rollup-stream.md
@@ -24,7 +24,7 @@ gulp.task('rollup', function() {
 
 ## Usage with sourcemaps
 ```js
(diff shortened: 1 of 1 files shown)


phated:
Thanks! Could you add a note about the version of rollup is required for this recipe to work?

hirusi:
@phated I have added the version to `@rollup/stream` while running `npm install`. `rollup` itself is a peer dependency so any changes to its version should reflect within the `@rollup/stream` package. I expect a breaking change in the API will be followed by a v2.x release so we should be covered with just an explicit version for the streaming package.

Let me know your thoughts! :slightly_smiling_face:

phated:
It seems that `@rollup/stream` is different than `rollup-stream` which was the previous recommended package. I've switched that reference and this looks great!

Thanks!

hirusi:
Oops - thanks for the catch. Yes, the rollup team had released a new package. :slightly_smiling_face:

PR #2467, opened 2020-07-26
add a copy function

SrStick:
a easy way to copy files or directories.


phated:
We don't extend the API very much, and everything is done in submodules

Issue #2256, opened 2018-10-25
[Docs] Advanced Section


Janiceilene:
We haven't written an advanced section as part of our documentation updates.

Some of the topics we want to cover are:
- [ ] Incremental builds
- [ ] Writing gulpfiles using another language
- [ ] Creating custom registries
- [ ] Writing a plugin
- [ ] Extglobs

This list may be added to as we come across more.

Feel free to suggest additional topics!

markstos:
Dynamically generated dependencies: A task starts, checks a flag or environment variable and decides which additional tasks needs to be run based on that.

markstos:
I found an example of dynamically generated dependencies here: https://github.com/gulpjs/gulp/blob/master/docs/getting-started/3-creating-tasks.md   Thanks.

donpedro:
I'd love to see "Writing an adapter", an analog of the invaluable [Writing a plugin](https://github.com/gulpjs/gulp/tree/master/docs/writing-a-plugin) docs. I might even be able to help with a first draft if I can get help with #2347 to figure out the questions I have on how to do it.

janiceilene:
Thanks for the suggestion @donpedro! I've added it to the OP. If you have get all your questions answered and want to work on a first draft of the docs, let us know!

donpedro:
Alright, I've get everything I needed to know figured out over on #2347. How can I help with docs?

I've got a bit of experience now with **Writing an Adapter**, but I know we're talking here about "Advanced Docs"; I'm not sure how advanced my knowledge is, I was thinking something more on the level of the existing docs for [Writing a plugin](https://github.com/gulpjs/gulp/tree/master/docs/writing-a-plugin). So if you have need for beginner-level docs maybe I can help.

Other things I could contribute if needed:
- **Writing gulpfiles using another language**: I'm writing everything in TypeScript (gulpfiles, plugins and adapters), so could talk about TypeScript specifically
- **Tooling**: We've got a pretty well-tuned setup to do live debugging in TypeScript (without transpiling) on gulp in vscode; I'm happy to do a writeup on that if you have a place for it

phated:
@donpedro Sorry for not following up here quickly, I've been traveling extensively and haven't had much time to dive in deeply.  I definitely think we can use some help on the documents you are suggesting. How would you like to contribute? We could set up a GoogleDoc or just interact on a PR—either works for us!

donpedro:
Great! Your call; either of those is fine by me!

[...]

PR #2479, opened 2020-08-15
Docs: Detail creation of custom registries

pixeldesu:
This PR picks up the "Creating custom registries" bullet point from #2256 

While writing this, I found that in the [undertaker](https://github.com/gulpjs/undertaker#registryregistryinstance) README, `registry()` is partially explained already, but I felt that just copy-pasting wasn't enough and the document here is a bit more thorough on what's exactly required or happening.

I also decided to link to all of the PoC registries that have been created in the organization beforehand, so people have some more starting points.

### DIFF
--- a/docs/advanced/creating-custom-registries.md
+++ b/docs/advanced/creating-custom-registries.md
@@ -11,64 +11,64 @@ Allows custom registries to be plugged into the task system, which can provide s
 
 ## Structure
 
-In order to be properly set up, custom registries need to follow a specific format to be accepted by Gulp.
+In order to be accepted by gulp, custom registries must follow a specific format.
 
 ```js
 // as a function
 function TestRegistry() {}
 
-TestRegistry.prototype.init = function (taker) {}
+TestRegistry.prototype.init = function (gulpInst) {}
 TestRegistry.prototype.get = function (name) {}
 TestRegistry.prototype.set = function (name, fn) {}
 TestRegistry.prototype.tasks = function () {}
 
 // as a class
 class TestRegistry {
-    init(taker) {}
+  init(gulpInst) {}
 
-    get(name) {}
+  get(name) {}
 
-    set(name, fn) {}
+  set(name, fn) {}
 
-    tasks() {}
+  tasks() {}
 }
 ```
 
-If a registry instance passed to `registry()` doesn't have these 4 methods available, it will throw an error.
+If a registry instance passed to `registry()` doesn't have all four methods, an error will be thrown.
 
 ## Registration
 
-If we want to register our example registry from above, we wil need to pass an instance of it to `registry()`.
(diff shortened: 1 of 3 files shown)


phated:
@pixeldesu WOW! This is **amazing**!!! Thank you for adapting the undertaker README for the wider gulp audience.

I'm wondering if you could adapt https://github.com/gulpjs/undertaker#sharing-tasks and https://github.com/gulpjs/undertaker#sharing-functionalities for the audience as well? Those seem really important for explaining how you'd perform those use cases with a custom registry.

Once we get this landed, I'll restructure it into an "advanced" section for our website (which will be our first!).

pixeldesu:
@phated Thanks for the feedback, didn't really expect it to be this well received! 😄 

I could definitely wrangle the two Undertaker sections into the document as well, I'm just thinking about a structure for that now.

Maybe a "Use Cases" section? I think the examples here are actually thorough enough, I would just adjust them to use Gulp/Gulp exports instead of Undertaker, of course.

So the final structure afterwards would be

**Creating Custom Registries**  
* ~Registry~ Structure (I guess I'll probably even drop the Registry here, it seems redundant, now looking at it)
* Registration
* Methods
* Use Cases
* Examples

phated:
Yeah, I think that is perfect actually!

What if the top 2 sections were flipped?
1. Registration (maybe use one of the examples as the import)
2. Creating a Registry (What do you think of this name?)
3. Methods
4. Use Cases
5. Examples

One thing I just remembered is that there are "not implemented" references to custom registries in the gulp documentation, could you update those links as well? That would help me have a good diff of where the references are.

pixeldesu:
> What if the top 2 sections were flipped?

Hmmm. Considering the topic is the creation of custom registries, first handling the structure, which is then registered makes a bit more sense in terms of order. I was conflicted about this myself, but registering already existing registries is already explained in the `registry()` docs pretty much.

> One thing I just remembered is that there are "not implemented" references to custom registries in the gulp documentation, could you update those links as well? That would help me have a good diff of where the references are.

I'll try to look for those and adjust them as necessary!

phated:
> Hmmm. Considering the topic is the creation of custom registries, first handling the structure, which is then registered makes a bit more sense in terms of order.

👍 Solid logic. I agree with your decision 😄

pixeldesu:
@phated done!

As mentioned, I took the examples as given, just adjusted them to be for Gulp specifically! Also, I only found a single reference to the page on the `registry()` API page itself and removed the reference, it should now be resolved over the front matter ID anyway.

If there's anything else, just mention it!

phated:
Thanks for all the work here @pixeldesu! This has been a great help 😄

phated:
This is live now! https://gulpjs.com/docs/en/advanced/creating-custom-registries

PR #2497, opened 2020-10-07
Remove some seemingly obsolete recipes

silvenon:
Remove the following recipes:

- selecting changed files
- running tasks in series
- exporting tasks
- running shell commands
- specifying a CWD
- splitting tasks in files
- using a config file

The explanations are in the commit messages.

This is of course very subjective and some removals may be too extreme. What do you think? This is in preparation for #2164. Also, does it make sense to still have Browserify recipes?


phated:
@silvenon perfect! I agree with all these removals and their reasonings! I will "rebase and merge". Thank you for following the conventional commit pattern, too!

PR #2499, opened 2020-10-10
Docs: Add recipe for static asset revisioning

silvenon:
Even though this was not requested in #2164, it's a common practice, so a recipe would be nice. 😃


docs/recipes/static-asset-revisioning.md
```diff
@@ -0,0 +1,33 @@
+# Static asset revisioning
+
+Static asset revisioning by appending content hash to filenames. Make sure to set the files to [never expire](http://developer.yahoo.com/performance/rules.html#expires) for this to have an effect.
+
+```sh
+npm install --save-dev gulp gulp-rev gulp-rev-rewrite gulp-rev-delete-original
```
phated, line 6: I don't want to recommend `gulp-rev-delete-original` because it's not maintained.

docs/recipes/static-asset-revisioning.md
```diff
@@ -0,0 +1,33 @@
...
+```js
+const gulp = require('gulp')
+const rev = require('gulp-rev')
+const revRewrite = require('gulp-rev-rewrite')
+const revDelOriginal = require('gulp-rev-delete-original')
+
+function revisionAssets() {
+  return gulp
+    // add assets you want to revision
+    .src(`dist/**/*.{css,js}`)
```
phated, line 18: Why would we generate the original files in `dist` just to remove them? We should show how to do this filtering in a pipeline instead of re-reading the files after output. Otherwise I don't think this recipe is useful.

phated (changes requested):
I'm not sure this recipe is useful. We shouldn't be showing people to write things to the filesystem, just to read them again and delete the originals.

Can you improve it?

silvenon:
How about this approach?

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
...
+const del = require('del')
+
+const isProd = process.env.NODE_ENV === 'production'
+
+function clean() {
+  return del('dist')
+}
+
+function styles() {
+  return gulp
```
phated, line 40: Let's use `pipeline()` (from node's Stream module) syntax any time things are being piped.

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
...
+  gulp.watch('styles/**/*.scss')
+  gulp.watch('views/**/*.njk')
+}
+
+// after everything is done, we no longer need the manifest file
+function deleteRevManifest() {
+  return del('dist/rev-manifest.json')
+}
+
+exports.dev = gulp.series(clean, gulp.parallel(styles, views), watch)
```
phated, line 75: I don't think we should have a separate `build` and `dev` script. You are already vary on `isProd` which should be enough.

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
...
+  const manifest = gulp.src(
+    'dist/rev-manifest.json',
+    // in development the manifest doesn't exist
+    { allowEmpty: !isProd }
+   )
+  return gulp
+    .src('views/**/*.njk')
+    .pipe(nunjucks.compile({ title: 'Hello world!' }))
+    // this updates the reference(s) to revisioned assets
+    .pipe(gulpIf(isProd, revRewrite({ manifest })))
```
phated, line 61: Wow, this `manifest` option is a terrible API that's not using gulp correctly. I think you should combine `views` and `styles` tasks into 1 pipeline and use gulp-filter to manage the "phases".

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
...
+
+function watch() {
+  gulp.watch('styles/**/*.scss')
+  gulp.watch('views/**/*.njk')
+}
+
+// after everything is done, we no longer need the manifest file
+function deleteRevManifest() {
+  return del('dist/rev-manifest.json')
+}
```
phated, line 73: If you combine the 2 tasks, we no longer need to delete some intermediary file on the filesystem.

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
...
+   )
+  return gulp
+    .src('views/**/*.njk')
+    .pipe(nunjucks.compile({ title: 'Hello world!' }))
+    // this updates the reference(s) to revisioned assets
+    .pipe(gulpIf(isProd, revRewrite({ manifest })))
+    .pipe(gulp.dest('dist'))
+}
+
+function watch() {
```
phated, line 65: I think we should just remove the watch stuff, it doesn't add anything to this recipe.

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
...
+const isProd = process.env.NODE_ENV === 'production'
+
+function clean() {
+  return del('dist')
+}
+
+function styles() {
+  return gulp
+    .src('styles/style.scss')
+    .pipe(sass().on('error', sass.logError))
```
phated, line 42: I'm wondering if this pattern even makes sense anymore with `pipeline()`

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
...
+const rev = require('gulp-rev')
+const revRewrite = require('gulp-rev-rewrite')
+const gulpIf = require('gulp-if')
+const del = require('del')
+
+const isProd = process.env.NODE_ENV === 'production'
+
+function clean() {
+  return del('dist')
+}
```
phated, line 37: If you are using static revisioning, I'm pretty sure you're never supposed to remove your old files (in case someone loads an old version from their cache).

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
+# Sass and Nunjucks with static asset revisioning
+
+Static asset revisioning by appending content hash to filenames. Make sure to set the files to [never expire](http://developer.yahoo.com/performance/rules.html#expires) for this to have an effect.
+
+Sass and Nunjucks are here just to demonstrate how to implement static asset revisioning in a more real-world scenario. The content has will be appended in production, so it's recommended to add the following npm scripts to `package.json`:
+
+```json
+{
+  "scripts": {
+    "dev": "gulp-dev",
```
phated, line 10: If we just implement 1 task, this can be `NODE_ENV=development gulp build`

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// after everything is done, we no longer need the manifest file
+function deleteRevManifest() {
+  return del('dist/rev-manifest.json')
+}
+
+exports.dev = gulp.series(clean, gulp.parallel(styles, views), watch)
+// in build it's important that "views" runs AFTER "styles"
+// if it runs before, the manifest file won't exist yet
```
phated, line 77: If we change to my suggestions above, you don't need this not anymore since everything will be streaming within one task.

phated (changes requested):
Better, but I think it still can be improved by not writing unnecessary steps to the filesystem.

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
silvenon: Could you be more specific? I never used it before.

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
silvenon: If we're removing watching (which I will), then it doesn't.

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
silvenon: Ok, I'll give it a shot.

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
silvenon: Also, nothing happens because I forgot to add tasks to them, so they are completely useless 😆

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
silvenon: I would like to avoid this, too, but, as I said, I need help. 🙏

docs/recipes/sass-and-nunjucks-with-static-asset-revisioning.md
silvenon: Oh… I thought that if an old version is being loaded, then static revisioning isn't set up correctly because the cache was supposed to be busted. That's a terrifying idea to me as an author that someone loads an old CSS or JS file, I guess things will break for them either way? 🤔

This clean task was not only for deleting old assets, but deleting old anything. Old pages etc. I want a 404 page to appear when a page has been deleted.  But I'll remove this task since we're focusing on revisioning only here.

silvenon:
I'm learning (more) about streams in Node.js now so I can finish this PR. I actually never studied them before, I somehow managed to get by in gulp without it, but not having this theory established is becoming pretty uncomfortable. 😄

silvenon:
If you have some good resources, I'm all ears. 👂

docs/recipes/static-asset-revisioning.md
mahdiamer: Good

PR #2503, opened 2020-10-18
Docs: Refactor Rollup recipe

silvenon:
Nothing extra is needed to use Rollup with gulp, so this refactors the
recipe to use Rollup's JavaScript API directly.


silvenon:
Alternatively, I could add this _next_ to the `rollup-stream` example, if you think it's useful to keep that one. But it's best to use tools directly unless it's really inconvenient.

phated:
I believe this recipe should describe both:
1. Rollup's normal JS API, which should be preferred if you don't need to apply other stream transforms
2. The new official stream API: https://www.npmjs.com/package/@rollup/stream if you do need to apply more stream transforms.

I think we can include both in this recipe.

phated:
@silvenon I also think this recipe should just be called `bundling`, since it will be the recommended way to include bundling inside your gulpfiles.

silvenon:
I started working on this again. My plan is to rename this to "Using Module Bundlers" instead because I also want to add webpack and esbuild, both of which support other bundling other assets, not just JavaScript. I'll finish Rollup in the following days and as separate PRs I'll add other bundlers as well.

phated:
@silvenon I think that is a great plan! Thank you for updating us.

phated:
@silvenon are you still planning to complete the above changes?

silvenon:
No, sorry, I lost my drive for some reason. 😔 I'll open a new PR once I get it back.

phated:
Thanks for letting me know! I'd like to re-open and pick this up myself. And thank you for the work you put in here already ❤️

PR #2545, opened 2021-02-20
Includes my documents

manoelmaciel:
A place to include private documentation.


phated:
This doesn't make much sense for the project.

manoelmaciel:
I agree ... Sorry. I followed an example and sent it by accident. I'll delete it!

PR #2572, opened 2021-05-06
Improve possibly confusing optional argument syntax

justsml:
The syntax for first-position optional parameters in function signatures can confuse newbies and devs from other languages.

The `gulp.task` API is so important that clarity & completeness should have a higher priority than EBNF notation.


phated:
I'm going to close this as I don't think it helps and actually recommends the non-recommended pattern.

PR #2592, opened 2021-06-22
Removed the gulp-sourcemaps plugin

brendan8c:
The gulp-sourcemaps plugin can be removed.
Gulp already has built-in support for source maps.
https://gulpjs.com/docs/en/api/src/#sourcemaps


phated (approved):
Excellent! Thanks!

PR #2642, opened 2021-11-08
Update README.md

TarunavBA:
Appveyor badge shows cancelled so i removed it!


phated:
We'll be normalizing the project coming up soon and applying the template is easiest if the stuff remains the same.

PR #2664, opened 2022-01-19
Revert "Docs: Remove gulp-sourcemaps because it is built-in (#2592)"

dnicolson:
This pull request reverts a previous change to the Browserify documentation relating to sourcemaps.

Unfortunately the previous change results in the following error:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Object
    at validateString (internal/validators.js:124:11)
    at Object.resolve (path.js:1039:7)
    at module.exports (node_modules/vinyl-source-stream/index.js:13:34)
```

### DIFF
--- a/docs/recipes/browserify-transforms.md
+++ b/docs/recipes/browserify-transforms.md
@@ -15,6 +15,7 @@ var source = require('vinyl-source-stream');
 var buffer = require('vinyl-buffer');
 var log = require('gulplog');
 var uglify = require('gulp-uglify');
+var sourcemaps = require('gulp-sourcemaps');
 var reactify = require('reactify');
 
 gulp.task('javascript', function () {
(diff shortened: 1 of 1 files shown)


phated:
You are half correct, vinyl-soure-stream won't take the option, but you can still write them with gulp. Want to fix that?

dnicolson:
Wouldn't it make more sense to do that in a different pull request? `gulp-sourcemaps` is still used in 7 other recipes.

phated:
As noted on the recipes (I forget exactly where), they are bad and probably should just be removed until people have a chance to write them better. They are only there as a reference for when people want to pick things up.

I think we either make them better or leave them alone.

dnicolson:
Reverting the breaking change and leaving it alone doesn't sound like a bad idea.

phated:
We don't do reverts, we only do fixes. I'm done arguing with you. If you want to fix the issue, feel free to send another PR.

PR #2690, opened 2022-07-11
Docs: Added missing semicolon

jv-k:


phated:
I don't believe that semicolon is necessary. We'll be adding prettier to the project to format all the docs in the future.

PR #2711, opened 2022-12-17
Updated Invalid Link [Docs | Stream-Handbook]

Viisheshjain05:
Link referenced to stream-handbook is wrong 
This is my first public PR hope this link is correct.

Updated Link - https://www.npmjs.com/package/stream-handbook

### DIFF
--- a/docs/writing-a-plugin/README.md
+++ b/docs/writing-a-plugin/README.md
@@ -193,7 +193,7 @@ if (someCondition) {
 
 If you're unfamiliar with streams, you will need to read up on them:
 
-* https://www.npmjs.com/package/stream-handbook (a MUST read)
+* https://www.freecodecamp.org/news/node-js-streams-everything-you-need-to-know-c9141306be93/
 * https://nodejs.org/api/stream.html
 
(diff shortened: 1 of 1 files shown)


phated:
We don't really want to link this to an npm package, especially one that substack is the only maintainer of... You probably want to use https://github.com/browserify/browserify-handbook

Viisheshjain05:
Reason for chosing npm package -
- Documentation - `How to write node programs with streams` which is provided in npm package.
- The github browserify is not appropriate link for that.

Alternative Options -
1. We can `add new .md` inside gulp docs providing better explaination for same.
2. We can use [Morioh stream-handbook](https://morioh.com/p/1666886ca798) this is the` same doc` on much trusted site.
3. `Remove that link` and add alternate links for explaining streams.
    - [freecodecamp - Everything on node.js streams](https://www.freecodecamp.org/news/node-js-streams-everything-you-need-to-know-c9141306be93/)
    - [nodesource - Understanding node.js streams](https://nodesource.com/blog/understanding-streams-in-nodejs/)


Thank you for commenting :smile:

phated:
Thanks for finding alternatives. I didn't fully review the freecodecamp article but it looks good enough.

phated (approved):

PR #2760, opened 2024-03-25
feat: Provide an ESM export

### LINKED ISSUE #2634: Support named exports in mjs files
Using `gulpfile.esm` I can do:

```js
import { src, dest, series, parallel } from "gulp";
```

When using `gulpfile.mjs`, this doesn't work because doesn't officially export the properties `src`, `dest`, etc.
Was wondering if there is any interest to support this, maybe in v5?

phated:
This implements an ESM wrapper following the Node.js docs for [Writing dual packages while avoiding or minimizing hazards](https://nodejs.org/docs/latest/api/packages.html#writing-dual-packages-while-avoiding-or-minimizing-hazards)

The test seems to work but I'm unsure if there is anything I need to do for the `bin` file. Perhaps someone else knows.

Closes #2634

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,26 +112,12 @@ exports.default = build;
 
 ## Use latest JavaScript version in your gulpfile
 
-__Most new versions of node support most features that Babel provides, except the `import`/`export` syntax. When only that syntax is desired, rename to `gulpfile.esm.js`, install the [esm][esm-module] module, and skip the Babel portion below.__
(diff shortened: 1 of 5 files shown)


perrin4869:
Since the `bin` field is never imported by the user, it just needs to follow the format specified by the root `package.json` file, no special care is needed 😃
Some people think it's good practice to also export the `package.json` file in the `exports` field

PR #2806, opened 2024-05-31
Update README.md to use deleteAsync

bramp:
For ~two years 'del' no longer exports a default method, and instead exports `deleteAsync()` and `deleteSync()`, the former replacing `del()`

See https://github.com/sindresorhus/del/commit/106d7d8ff813059c437d7c428539b7719292d1d6


yocontra:
This seems like a good opportunity to eliminate a dependency and use `import fs from 'fs/promises'` + `rmdir` since this example was written before fs/promises existed del was the cleanest to use

PR #2831, opened 2025-02-18
Update 4-async-completion.md

IOLOII:
Starting with RxJS 6.x, the Observable.of method has been removed, and it's recommended to directly import the of function from 'rxjs' instead.


IOLOII:
```javascript
const { Observable } = require('rxjs');
function observableTask() {
  return Observable.of(1, 2, 3); // TypeError: Observable.of is not a function
}
exports.default = observableTask;
```

PR #2835, opened 2025-06-01
chore: Fix CI

phated:
Figuring out the CI bitrot.


phated:
Streamx is broken on Node 10. See https://github.com/mafintosh/streamx/issues/101

PR #2846, opened 2026-04-20
docs(watch): include glob in documented watch task error message

Bojun-Vvibe:
**Repo:** gulpjs/gulp (⭐ 33000)
**Type:** docs
**Files changed:** 1
**Lines:** +1/-1

## What
Updates `docs/api/watch.md` to accurately describe the error message thrown when a string or array is passed as the `task` argument to `gulp.watch()`. The actual error message includes a `"watching <glob>: "` prefix that names the offending glob, but the documentation omitted it.

## Why
The Errors section of the `watch()` API doc claims the message is `"watch task has to be a function (optionally generated by using gulp.parallel or gulp.series)"`, while the implementation in `index.js` (and the corresponding tests in `test/watch.js`) actually throws `"watching <glob>: watch task has to be a function..."`. Users grepping logs or writing assertions against the documented string would be misled. This aligns the docs with real runtime behavior.

## Testing
Verified the actual thrown message in `index.js:30-32` and the existing tests in `test/watch.js:231-255` that assert on it. Pure docs change — no code paths affected, no tests need updating.

## Risk
Low — docs-only change to a single sentence; no behavior change.


phated:
We are not interested in your slop.

